PR #3, opened 2016-10-25
Allows setting DATE values for DTSTART and DUE

korelstar:
continuing from owncloud/tasks#356:

> Fixes owncloud/tasks#18.
> 
> According to [RFC 5545](https://tools.ietf.org/html/rfc5545), `DTSTART` and `DUE` can be either of type `DATETIME` or `DATE`. Until now, this is not respected by the GUI which forces `DATETIME`. With this pull request, times and the time input field are hidden, if the type of `DTSTART` or `DUE` is `DATE`. Furthermore, a new checkbox allows to switch between `DATETIME` and `DATE`. Since according to RFC 5545, `DTSTART` and `DUE` must have the same type, this change applies to both fields, if set.
> 
> oc-tasks-allday

### DIFF
--- a/js/app/app.js
+++ b/js/app/app.js
@@ -76,6 +76,16 @@ angular.module('Tasks').run([
 				sameElse: '[' + t('tasks', 'Due on') + '] MMM DD, YYYY, HH:mm'
 			}
 		});
+ 		moment.locale('details_allday', {
+ 			calendar: {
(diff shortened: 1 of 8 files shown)


raimund-schluessler:
Works as expected. 👍 
@korelstar Thanks for the work on this 😃

jancborchardt:
Awesome stuff @korelstar! :)

PR #10, opened 2016-11-10
Fix occasionally defective date when using allday

korelstar:
When using the new all-day feature (#3), sometimes the saved date was defective. Then, the frontend looks correct, but in the backend iCal entry, the date was saved in the format "YYYY-MM-" instead of "YYYYMMDD". After refreshing the tasks app, a wrong date was shown.

I've fixed this bug by introducing a new helper method `momentToICALTime` which uses the respective `ical.js` functions in order to create a `date` or `datetime`.


raimund-schluessler:
I can confirm the problem. It occured when all-day was enabled and afterwards the date was changed. The fix solves the problem.

@korelstar Thanks for fixing this.

PR #11, opened 2016-11-14
Ensure that START < DUE if both set

korelstar:
Fixes owncloud/tasks#351: The standard forbids that START>DUE, so these changes have the purpose to be conform to the respective [RFC 5545](https://tools.ietf.org/html/rfc5545#section-3.8.2.3).

- [x] default values for `START` and `DUE` (i.e. if the field was unset before and should now be set) are calculated from `DUE` (if `START` should be set and `DUE` is set and is in the past) resp. `START` (if `DUE` should be set and `START` is set and is in the future) instead from the current time.
- [x] default value for `START` is now rounded downwards (instead of upwards) in order to calculate a value with `START<DUE`
- [x] if `START` or `DUE` is changed so that `START>DUE` would be true, the other value (`DUE`/`START`) will be changed, so that the time between `START` and `DUE` will remain like before the change. Therefore, if `START<=DUE` before changing the task, then `START<=DUE` afterwards.
- [x] disallow `START=DUE`, see the discussion under owncloud/tasks#351

### DIFF
--- a/js/app/services/businesslayer/tasksbusinesslayer.js
+++ b/js/app/services/businesslayer/tasksbusinesslayer.js
@@ -209,8 +209,10 @@ angular.module('Tasks').factory('TasksBusinessLayer', [
 				var start = moment(task.start, "YYYY-MM-DDTHH:mm:ss");
 				var due = moment(task.due, "YYYY-MM-DDTHH:mm:ss");
 				if (!start.isValid()) {
-					var reference = due.isBefore() ? due : moment();
-					reference.subtract(1, 'm');
+					var reference = moment().add(1, 'h');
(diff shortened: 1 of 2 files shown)


korelstar:
I've implemented the last missing task: disallow `START=DUE`, see the discussion under owncloud/tasks#351. From my side, it's ready to merge.

raimund-schluessler:
I will have a look.

raimund-schluessler:
@korelstar The PR looks fine. I just made a small change, so that the start date is in the future (next full hour), except it would be ```START>DUE```.

If you are fine with this change, feel free to merge.

PR #51, opened 2017-02-18
[WIP] Fit to styleguide, closes #49

### LINKED ISSUE #49: Deletion countdown cancelation not working since the last popover unification
Since nextcloud/server#3024

- [ ] Clicking the cancel button still delete the task
- [ ] Design is broken
- [ ] Does not fit the guidelines ⚠️

raimund-schluessler:
Closes #49

### DIFF
--- a/templates/main.php
+++ b/templates/main.php
@@ -31,13 +31,11 @@
                     </span>
                     <span class="title">{{ collection.displayname }}</span>
                 </a>
-                <div class="app-navigation-entry-utils">
-                    <ul>
-                        <li class="app-navigation-entry-utils-counter">{{ getCollectionString(collection.id) | counterFormatter }}</li>
-                    </ul>
+                <div class="app-navigation-entry-utils-counter">
+                    <span>{{ getCollectionString(collection.id) | counterFormatter }}</span>
                 </div>
             </li>
-            <li class="list with-menu handler"
+            <li class="list handler"
                 id="list_{{ calendar.uri }}"
                 calendarID="{{calendar.uri}}"
                 ng-repeat="calendar in calendars"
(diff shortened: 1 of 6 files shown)


raimund-schluessler:
I fixed all issues I found and tried to fit the guidelines as far as possible. The only thing left is the positioning of the popover menu under the three-dots-icon which is slightly off.

@skjnldsv Could you maybe have a look at this? I thought the popover menu would be positioned automatically when it is located within the div containing the three-dots-icon?

templates/main.php
```diff
@@ -31,13 +31,11 @@ class="collection"
                     </span>
                     <span class="title">{{ collection.displayname }}</span>
                 </a>
-                <div class="app-navigation-entry-utils">
-                    <ul>
-                        <li class="app-navigation-entry-utils-counter">{{ getCollectionString(collection.id) | counterFormatter }}</li>
-                    </ul>
+                <div class="app-navigation-entry-utils-counter">
```
skjnldsv: You need the   ` <div class="app-navigation-entry-utils">` to fit the css guidelines.
https://docs.nextcloud.com/server/11/developer_manual/app/css.html#menus

templates/main.php
```diff
@@ -51,35 +49,38 @@ class="collection"
...
-                        <li>
-                            <a href="{{calendar.exportUrl}}" download="{{calendar.uri}}.ics">
-                                <img class="icon-download svg" src="<?php p(image_path('core', 'actions/download.svg'))?>"/>
-                                <span><?php p($l->t('Download')); ?></span>
-                            </a>
-                        </li>
-                        <li confirmation="delete(calendar)" confirmation-message="deleteMessage(calendar)">
-                        </li>
-                    </ul>
+                <div class="app-navigation-entry-utils-counter">
```
skjnldsv: Should be first. Before the `app-navigation-entry-menu` and the `app-navigation-entry-utils-menu-button` button.

And it's empty on my installation, therefore moving the menu icon to the left. 
capture d ecran_2017-02-20_05-36-31

skjnldsv:
Nice work! Still a shame that you need to add your own css to the popover menu for the countdown to work. But the countdown works fine now. 😃

templates/main.php
raimund-schluessler: Ah, alright, thanks for the manual link. Would it be allowed to have the counter on the second position to it doesn't get shifted when the three-dots-icon is visible?

templates/main.php
raimund-schluessler: Yeah, I notived that too. I will align the thee-dots-icon to the right in case no counter is visible.

skjnldsv: No need to do that, it should align itself if at the right place :)

raimund-schluessler: Yes, but see my comment above. I would like to switch the places of counter and three-dots-icon, but I guess this is against the guidelines !?

skjnldsv: Sorry, I thought you said that you will align it, not that you wanted to :)
Well, I don't think this is against the guidelines, but if you hide the counter properly, the menu should align automatically on the right :)

skjnldsv: I will take a look asap 😉

skjnldsv:
@raimund-schluessler I pushed a pr to fit the css guidelines.
The only thing to fix is the menu now. :)

skjnldsv:
Okay, I looked at the overall changes you did and found some strange stuff.
Why did you change the menu/menubutton div order? The app was fitting the css guidelines before, the popovermenu content didn't. There was no needs for that didn't it?

raimund-schluessler:
Thanks for fixing this stuff. I haven't checked how it looks now, but I will do so soon.

> Why did you change the menu/menubutton div order? The app was fitting the css guidelines before, the popovermenu content didn't. There was no needs for that didn't it?

What exactly do you mean? Could you give a link to the line in the code?

skjnldsv:
Well, if you look at my pr and the overall changes, I pretty much rolled back to the original code (without knowing it). That means the structure of the app-navigation was good on the first place. So why did you changed it? The only stuff that was necessary to change was the popover menu with the countdown, why editing the rest?

raimund-schluessler:
Yeah, you are right. I guess I screwed that up in an attempt to fix the positioning of the three-dots-icon without realising. Thanks for reverting that.

But there are some new issues now.

- [ ] The counter is not always visible, only when the list is active. It should always be visible
- [ ] The counter now shows ```0``` when no tasks are present, whereas it simply showed nothing before. I know, it counterFormatter in the guidelines does the same, but I think it clutters the layout with a useless display.

I will fix the first issue and would like some feedback on the second. Also @jancborchardt for the second point.

raimund-schluessler:
Wait, you moved the popover menu again outside of the div containing the three-dots-icon. But in https://github.com/nextcloud/server/issues/2798#issue-196891164 you say

> You need to put the entire menu just after the three dot icon (refs #2545 (comment))
>
>    ```<div><span class="icon-more"></span><div class="popover">...</div></div>```
>    With this, we don't need js, css only is ok.

What is correct now?

skjnldsv:
Okay, I understand the confusion now! :)
I wrote the guide for the popover menu only. Not for the app-navigation menu.

The css guidelines on the dev manual still apply there. The only stuff that changed which is the same for the two menus is the internal structure of elements. (span, divs...)

raimund-schluessler:
Ah, ok, thanks for the insight 😄 

I will fix the issues then.

skjnldsv:
We should probably revert all of this pr and only fir the countdown ^^'

Sorry about the confusion, the dev manual not beiing updated is pretty annoying (don't have time for that yet 😢 )

raimund-schluessler:
Well, the countdown is not the only thing that had to be changed. There actually were several issues including opacity, the sortorder dropdown and the input field.

But starting a new PR and applying these fixes there might be an option.

raimund-schluessler:
I will split the necessary changes up into separate PRs.

Fix input fields: https://github.com/nextcloud/tasks/pull/52

jancborchardt:
Great work on fitting the app to the styleguide! :)

PR #52, opened 2017-02-20
Fix task input fields

raimund-schluessler:
Fixes the task input fields


raimund-schluessler:
I am going to merge this since the changes are minor.

PR #53, opened 2017-02-20
Fix app navigation menu, closes #49

### LINKED ISSUE #49: Deletion countdown cancelation not working since the last popover unification

raimund-schluessler:
@skjnldsv Could you have a look here, please? I fixed the app navigation menu including the delete countdown.

<strike>The sortorder dropdown list needs some attention now, but I will do this in a follow up PR.</strike>

I decided to fix the sortorder dropdown here too. Should be all good now.

### DIFF
--- a/templates/main.php
+++ b/templates/main.php
@@ -59,20 +59,20 @@
                 <div class="app-navigation-entry-menu" ng-show="calendar.writable">
                     <ul>
                         <li>
-                            <span ng-click="startEdit(calendar)">
-                                <img class="icon-rename svg" src="<?php p(image_path('core', 'actions/rename.svg'))?>"/>
+                            <a ng-click="startEdit(calendar)">
+                                <span class="icon-rename svg"></span>
                                 <span><?php p($l->t('Edit')); ?></span>
-                            </span>
+                            </a>
                         </li>
                         <li>
-                            <span ng-click="showCalDAVUrl(calendar)">
-                                <img class="icon-public svg" src="<?php p(image_path('core', 'actions/public.svg'))?>"/>
+                            <a ng-click="showCalDAVUrl(calendar)">
+                                <span class="icon-public svg"></span>
                                 <span><?php p($l->t('Link')); ?></span>
-                            </span>
+                            </a>
                         </li>
                         <li>
                             <a href="{{calendar.exportUrl}}" download="{{calendar.uri}}.ics">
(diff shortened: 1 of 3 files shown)


raimund-schluessler:
@skjnldsv would be awesome, if you could have a look :smile:

templates/main.php
```diff
@@ -167,46 +167,46 @@ class="collection"
             <div class="app-navigation-entry-menu bubble sortorder-dropdown">
```
skjnldsv: This is NOT a `app-navigation-entry-menu`, we're in the content, this should be a popovermenu 😉 
Though this should roughly be similar, changing the class should be enough :)

The menu content on the other hand, seems to be broken (triangle), so since we're working on this, maybe fix this too? 😄

skjnldsv: If you want to work on this in another pr, I will change my review to approved to merge this fast! ;)

raimund-schluessler: I created an issue: https://github.com/nextcloud/tasks/issues/57

skjnldsv: The click on menu is broken too, this is not just the guidelines 😃 
When you select a sort option, you can't click the button again (or with a lot of difficulties)

raimund-schluessler:
Great, thanks for reviewing. Will tackle the popover problem asap.

templates/main.php
raimund-schluessler: Hm, I never considered this a bug. But I am ok with keeping the menu open after clicking the popover.

skjnldsv: No no, Im talking about the button, not the menu :)

raimund-schluessler: Sorry, I don't get what you mean. Clicking on the sort-order dropdown button opens the menu, clicking it again, closes it. When I click inside the popover, a sorttype is selected and the menu is closed. What is wrong about that? 😕

skjnldsv: Ahaha, sorry, you're right I'm not clear!
When you click the button, the popover opens, you can select an option and it closes right away, which is perfect!
BUT, when after that you try to click te button again, it doesnt work. You need to click just under the icon. It doesn't work after that.

raimund-schluessler: Hm, I can't reproduce this behavior. After selecting an option I can open and close the menu just fine. It behaves the same as before. Could you maybe create an issue for that and tell me which browser you use?

PR #54, opened 2017-02-20
Fix button border radius

raimund-schluessler:


raimund-schluessler:
I am going to merge this since the changes are minor.

PR #66, opened 2017-03-15
Add some much needed aria attributes and make the app at least usable by screen reader users.

tyrylu:
Due to the count of custom controls like the completed and all day checkboxes, the usability for visually impaired users was poor, to say the least.
This pull request fixes the most obvious stuff:
- Adds the role="checkbox" and aria-checked to the completed task checkbox and to the all day checkbox in the date picker for the end date
- Adds some aria labels to the icons, but some ui designer should probably look in to those because it was not entirely obvious what is an useful clickable element
- Tries to convey the completeness state in the list, but no screen reader reads aria-labels or anything i came up with on an empty div


raimund-schluessler:
Thanks for this contribution. I wasn't aware of that and I must admit that I have no idea how it works. Although I am not sure how to test this, I am glad to merge this PR as long as it doesn't affect the "normal" usage (which it seems it does not).

Is this PR work in progress or should we already review it? Also, could you please squash the three commits into a single one (or I can do it, once you consider this PR ready for merging)?

tyrylu:
Well, this definitely does not fix all of the issues, but as i said, it at least fixes some of the more obvious ones. As to the testing, try to use the tasks app with NVDA (http://nvda-project.org) and see for yourself how it turns out. :-) So, because having something is better than nothing, i would rather merge it now and continue with the other unlabeled icons/pseudo-buttons/what they are later.
And i would appreciate the squashing, by the way and thanks for it.

PR #71, opened 2017-05-02
Enhance input fields which looks like disabled

Dennis1993:
I think the Input field for new entries Looks like disabled Input fields. The same on the right side if you add a comment. I removed the grey Background color to enhance the visibility of editable input fields.

Before:
before

After:
after

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -567,7 +567,6 @@
   font-size: 13px;
   width: 100%;
   color: #505050;
-  background: none repeat scroll 0 0 transparent !important;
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
I am fine with the changes. What does @jancborchardt say?

raimund-schluessler:
Ok, lets merge this.

jancborchardt:
Awesome work @Dennis1993, good stuff! :) Sorry for my late reply.

PR #73, opened 2017-05-07
Fix slider when narrow and adjustment for nc12

### LINKED ISSUE #68: Can't change priority with slider when window is to narrow
### Steps
1. Create a task
2. Open the sidebar
3. Open the priority option
4. Reduce the window size until the slider falls into a new line:
> aufgaben_-_jobisch_cloud_-_2017-03-28_10 55 21
5. Try to change the priority with the slider
6. See how the focus changes to "% done"

@raimund-schluessler

mkuhlmann:
Fixes #68. Also input was too large for the row since nc12 introduced `min-height: 32px;` on all inputs.

Before: 
localhost-index php-apps-tasks- nexus 5x 1
After:
localhost-index php-apps-tasks- nexus 5x 2

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -934,12 +934,13 @@
 #app-sidebar div.content-wrapper .body .section input[type="range"] {
   border: medium none;
   box-shadow: none;
-  width: 190px;
-  margin: 5px 36px 0 0;
+  width: calc(100% - 90px);
+  margin: 5px 36px 0 5px;
   height: 17px;
 }
 #app-sidebar div.content-wrapper .body .section input[type="text"] {
   float: left;
+  min-height: 0;
 }
 #app-sidebar div.content-wrapper .body .section .select2-search-field input {
   border: 0 none;


raimund-schluessler:
Thanks for fixing this, I will have a look.

css/style.css
```diff
@@ -934,12 +934,13 @@
 #app-sidebar div.content-wrapper .body .section input[type="range"] {
   border: medium none;
   box-shadow: none;
-  width: 190px;
-  margin: 5px 36px 0 0;
+  width: calc(100% - 90px);
+  margin: 0 36px 0 4px;
```
nickvergessen: I'd keep the 5px for the first position, looks a bit more inline then.

nickvergessen:
Fixes #68 as advertised 👍

css/style.css
mkuhlmann: Looks like this for me with 5px 36px 0 4px then:
localhost-index php-apps-tasks- nexus 5x

nickvergessen: I tested with firefox and chromium on my laptop and both need the 5px

mkuhlmann: That's strange. Changed it like you suggested ;)

Issue #101, opened 2017-08-28
Update build environment


raimund-schluessler:
- [x] Use Yarn instead of NPM
- [x] Use gulp instead of grunt
  - [x] Minify js files
  - [x] Lint js files
- [x] Use Yarn + gulp instead of bower

Requires #56 and #95

raimund-schluessler:
Remaining SCSS issue will be tackled in https://github.com/nextcloud/tasks/issues/95

PR #105, opened 2017-09-02
Update Build environment

raimund-schluessler:
Implements #101

- [x] Use Yarn instead of NPM
- [x] Use gulp instead of grunt
  - [x] Minify js files
- [x] Use Yarn + gulp instead of bower

#94 Should be in first.


raimund-schluessler:
There is a problem with using Yarn + Gulp instead of bower:

The jquery-timepicker package https://github.com/fgelinas/timepicker which is currently used is not available via NPM since it does not have a `package.json` in its git repository. Since the last commit there was nearly 4 years ago, I suppose this is not going to change.

For now, I will manually checkin the old timepicker in a `vendor_legacy` folder, which is not maintained by NPM.

In the long run, we should move to a still maintained timepicker script, maybe directly in server: https://github.com/nextcloud/server/issues/6348

skjnldsv:
Build works fine! :)

raimund-schluessler:
> But why is there so many commits?
> Why not add only the required ones? I can't keep up with all of this! :'(

I am sorry for this. The problem is that they are hard to separate at the moment because they depend on what was done in the previous PRs (which are not merged yet). We should start to merge with #94 and then I can rebase and they will be a lot better to review.

skjnldsv:
@raimund-schluessler Is this the next one? :)

raimund-schluessler:
Yes, I am rebasing at the moment.

raimund-schluessler:
Ok, it is rebased and the build still works as it should. I will merge this one and go on with rebasing #107.

skjnldsv:
` Total time 2 min 28 sec`
Wow that's fast! :D 🚀

Issue #95, opened 2017-07-26
Cleanup CSS code and move to SCSS


raimund-schluessler:
Also lint SCSS files and compile them into a single file.

PR #107, opened 2017-09-03
Lint javascript files

raimund-schluessler:
Needs #94 and #105 first.


skjnldsv:
It's like a treasure map! Each clue found, we move towards the next pr!!  🎠

raimund-schluessler:
> It's like a treasure map! Each clue found, we move towards the next pr!! 🎠

With #95 and #105 merged I will now rebase #108 onto master.

raimund-schluessler:
I will merge this, since it is only cleanup with no functional changes. Having this in makes the whole development a lot easier.

PR #108, opened 2017-09-04
Move to scss, closes #95

### LINKED ISSUE #95: Cleanup CSS code and move to SCSS

raimund-schluessler:
Still needs some work:

- [x] Move to scss
  - [x] Lint scss files
  - [x] Concat scss files
- [x] Fix icons without icon
- [x] Cleanup scss code (especially the colors), check server styles
- [x] Fix manual `background-position` for `ico-*`-styles
- [x] Change icon prefix to `icon-*`
- [x] More cleanup


raimund-schluessler:
I changed the icon-prefix to `icon-*`. This required quite some work, but at least it is possible at all using SCSS. I had to rename the files `star-*` because otherwise this core rule https://github.com/nextcloud/server/blob/master/core/css/icons.scss#L358-L362 would have overwritten the `background-position` on hover.

raimund-schluessler:
Also, for some weird reason the SCSS engine only allows `@extend: .rule` if there is only a single `.rule` in the whole file.
If there is both something nested like `#id .rule` as well as `.rule` in the same file it will break the compilation process. I would have expected that to be unambiguous, but it is not?!
@skjnldsv Is this expected, because not valid SCSS, or is it a bug?

skjnldsv:
I think the extend is limited because the apps use their own file compiler. Since none of the core scss are used when compiling it, aside from variables.scss, it can't extend a core property! :(

skjnldsv:
Oh I think I misread. Indeed nested selectors can't be extend in sass if my memory is correct! :)

raimund-schluessler:
Hm, I was aware, that nested selectors cant be extended, but I thought this would work, since I especially did not want to use the nested selector, but the non-nested root one.

However, `@extend` seems to work different than I thought.
Using http://beautifytools.com/scss-compiler.php, for this SCSS code:
```scss
// SCSS source code
.rule {
	color: #ff0000;
}

.ext-rule {
	@extend .rule;

	background-color: #00ff00;
}

#id123 .rule {
	color: #0000ff;
}
```
I would have expected to get something like this:
```css
// CSS expected result
.rule, .ext-rule {
	color: #ff0000;
}

.ext-rule {
	background-color: #00ff00;
}

#id123 .rule {
	color: #0000ff;
}
```
but instead I get this:
```css
// CSS real result
.rule, .ext-rule {
	color: #ff0000;
}

.ext-rule {
	background-color: #00ff00;
}

#id123 .rule, #id123 .ext-rule {
	color: #0000ff;
}
```

In server, the above SCSS code does not compile at all.

raimund-schluessler:
Also, if I have something like this
```scss
.rule {
	color: #ff0000;
}

.ext-rule {
	@extend .rul;

	background-color: #00ff00;
}
```
in my `style.scss`, while I expect this to not compile (.rul is not defined), it breaks the whole instance, giving an internal server error on plain white background.

raimund-schluessler:
I think this is pretty much done. @pixelipo @skjnldsv Mind having a look?

It might be good to have the PRs #94, #105 and #107 merged before reviewing this one here, so it gets a bit more clearly.

skjnldsv:
Of course!! :)
I will asap!!

skjnldsv:
I can't review. There is far too much commits in there, I don't know which is relevant to this pr or not! :(

raimund-schluessler:
> I can't review. There is far too much commits in there, I don't know which is relevant to this pr or not! :(

Yes, I was worried about that already. The commit starting from Sep 4, 2017 belong to this PR, the ones before belong to #94, #105 and #107. The problem is, that the move to SCSS does not work / doesn't make sense without them.

It is probably best if we merge the other PRs first, then I rebase this one and then you can review.

skjnldsv:
That's not really the best way to use differents pr.
You could create a single branch which you rebase on local for your test alone, but for the sake of github and colaborative work, you can't have 4 pr with 20000+ lines and the same commits! :(

Reviewing others will be as painful as this one. I'm sorry to say that! :/
I will definitely help with the scss and stuff once there will be only commits related to this pr purpose. :)

raimund-schluessler:
Well, I normally don't do that, but how am I supposed to move to SCSS when the build environment needs to be changed for that and the respective PR is not available in master yet?

Anyway, I will ping you once this is resolved.

skjnldsv:
@raimund-schluessler the scss conversion is core-related, it should work instantly.
You just need to rename it to .scss and convert it :)

You could even cleanup things as it's n ot supposed to change or break stuff. 😉 
I'm working on reviewing the 94! 👍

raimund-schluessler:
> @raimund-schluessler the scss conversion is core-related, it should work instantly.
> You just need to rename it to .scss and convert it :)

That's true in principle. But the move to svg_sprite made a lot of changes to `styles.css` necessary which I then would have to implement again on the SCSS file. I picked what I believed to be the better of two bad options ;)

> I'm working on reviewing the 94! 👍
You want me to rebase #94 first? Should have done that in the first place....

skjnldsv:
> That's true in principle. But the move to svg_sprite made a lot of changes to styles.css necessary which I then would have to implement again on the SCSS file. I picked what I believed to be the better of two bad options ;)

Well, scss is retro compatible with scss, so you could have put it at the end of the file and evolve it to scss after? 😉 

> You want me to rebase #94 first? Should have done that in the first place....

Don't bother :)

raimund-schluessler:
@skjnldsv Ok, this is finally rebased. The PR is still quite big, but only contains stuff related to SCSS now.

skjnldsv:
You're awesome!! Let's review before going to bed!

skjnldsv:
@raimund-schluessler what's the diff between the scss in /css/src and the other one in /css?

raimund-schluessler:
> @raimund-schluessler what's the diff between the scss in /css/src and the other one in /css?

I use gulp to compile the scss files (sprites and style) in `/css/src` into a single scss file so that I have only one css file to load. I could rename them to make this clear if you want.

skjnldsv:
It could, but it should fit your needs, not mines! :)
I will post my comments on the src ones then! Thanks!  😉

raimund-schluessler:
> It could, but it should fit your needs, not mines! :)

I don't mind the naming convention (also, I didn't think much about it 😉). If we can prevent confusion by renaming, I am totally fine with it.

Also, thanks a lot for reviewing. 👍

css/src/style.scss
```diff
@@ -0,0 +1,1703 @@
+//colors
+$red: #f00;								// warnings and errors
```
skjnldsv: You should use the core variables
https://github.com/nextcloud/server/blob/da5662c454cbdf9e81c6ac83cc0dfe72692fd883/core/css/variables.scss#L1-L8

css/src/style.scss
```diff
@@ -0,0 +1,1703 @@
...
+		max-height: 100px;
+
+		> a {
+			padding-right: 20px;
+		}
+
+		&.active {
+			> a span {
+				&.title {
+					opacity: 1;
```
skjnldsv: Not needed, the opacity is already on 1 on active

css/src/style.scss
```diff
@@ -0,0 +1,1703 @@
...
+				margin-left: 2px;
+			}
+
+			&.title {
+				margin-left: 11px;
+			}
+		}
+	}
+
+	.app-navigation-entry-edit {
```
skjnldsv: You could merge properties from lines 210-224

css/src/style.scss
```diff
@@ -0,0 +1,1703 @@
...
+	.label {
+		margin-right: 10px;
+	}
+
+	.active .icon.sort-indicator {
+		visibility: visible;
+	}
+}
+
+.app-navigation-entry-menu {
```
skjnldsv: Could be merged with some of the prperties of the line 292

skjnldsv:
@raimund-schluessler I'm so sorry if i sound condescending or rude, it's absolutely not my intention. i just gave my first impression as someone who just came in the task app after only reading the original css file! :)

You seemed to have made quite an impressive job here, and this is awesome! I just don't have the time to rephrase everything to ease the edge! :)
I'm sorry! I will gladely help you all I can if you want my help though!! 😉

raimund-schluessler:
> @raimund-schluessler I'm so sorry if i sound condescending or rude, it's absolutely not my intention. i just gave my first impression as someone who just came in the task app after only reading the original css file! :)

Don't worry, it's fine. I can handle criticism. And your comments are all relevant and helpful. 👍 

I indeed used a converter, but I guess the source material / the css file was already quite bad and disorganized. I am not quite sure how to proceed here. I will implement the changes you proposed, but I guess the `css/src/styles.scss` file needs a whole makeover, more or less from scratch (as you proposed).

I think about whether or not we should merge this PR anyway and proceed in a follow-up PR to cleanup the SCSS file? What do you think?

skjnldsv:
I would suggest not to merge.
I can provide a good way to restart the makeover if you want to! :)

raimund-schluessler:
> I would suggest not to merge.
I can provide a good way to restart the makeover if you want to! :)

Ok, let me just rebase the PR, to have it a bit more tidy before starting the makeover 😉

raimund-schluessler:
Ok, not super tidy, but it should do it.

raimund-schluessler:
I splitted the scss file into separate parts and tried to sort it a bit. Next step is to clean up each file.

skjnldsv:
Okay, good idea! We should get rid of the variables too!
We should have enough vars on the server scss to use for tasks. :)

raimund-schluessler:
Good idea, but you need to recompile the style.scss file afterwards. Either with gulp watch or just gulp. 😉

skjnldsv:
Will do it after I finished! :p

raimund-schluessler:
I rebased again onto current master. You should run `gulp watch` inside the js folder, since it will lint the scss and compile it into the styles.scss file. Otherwise you won't see the changes in the browser ;)

pixelipo:
@raimund-schluessler sorry I couldn't properly review #94 , but this PR seems to handle most of the issues. I will test it tomorrow unless you merge before :wink:

skjnldsv:
It's still in progress! :)

raimund-schluessler:
> @raimund-schluessler sorry I couldn't properly review #94 , but this PR seems to handle most of the issues. I will test it tomorrow unless you merge before 😉

Don't worry, I am glad you don't have any major concerns about #94. We more or less had to merge it, in order to go on with the later PRs. This PR here is very much WIP still.

skjnldsv:
@raimund-schluessler do you still needs help on here? :)

raimund-schluessler:
@skjnldsv I currently work my way through the css and clean everything up. I wouldn't mind if you could have a look as well ;)

raimund-schluessler:
- [ ] The select2 in the sidebar needs cleanup and fixing

raimund-schluessler:
@skjnldsv You need to compile your changes, otherwise you won't see the effect. The app now looks like this: 🙈 

bildschirmfoto am 2017-09-25 um 09 03 10

raimund-schluessler:
Also, I would not replace ```$red_overdue```. This color is matched to the red icons, and having the icons colored ```#ff0000``` looks weird.

css/src/06sidebar.scss
```diff
@@ -214,7 +214,7 @@
 				}
 
 				&.high .section-title {
-					color: $red_overdue;
+					color: $color-error;
```
raimund-schluessler: I would not replace ```$red_overdue```. This color is matched to the red icons, and having the icons colored ```#ff0000``` looks weird.

css/src/05taskbody.scss
```diff
@@ -70,19 +70,19 @@
...
-		color: $blue_due;
+		color: $color-primary;
 		float: right;
 		font-size: 11px;
 		line-height: 14px;
 		margin: 10px;
 
 		&.overdue {
-			color: $red_overdue;
+			color: $color-error;
```
raimund-schluessler: Same here.

css/src/05taskbody.scss
```diff
@@ -70,19 +70,19 @@
 	}
 
 	.duedate {
-		color: $blue_due;
+		color: $color-primary;
```
raimund-schluessler: Same goes for ```$blue_due```, replacing it with ```$color_primary``` does not make sense, since it is matched to the icon color.

raimund-schluessler:
@skjnldsv Please have a look at my comments regarding the colors 😃

skjnldsv:
@raimund-schluessler Thank you! I did some fixes only looking at the code. Good catch! :)

raimund-schluessler:
@skjnldsv I thought about the variables thing. Is there any particular reason, why you don't want to use variables for the colors here? I think they would increase maintainability a lot.

Without variables now, we have 60+ different locations where to check for the color. If I want to change one color now, I need to check on 10 or more lines and must not forget to change every line where this particular color is used. With variables there would only be one line, and it would be certain that only the limited amount of colors defined in variables.scss is used.

If there is no special reason as to why not use variables here, I would vote for reversing the last three commits and put the ```nc-lighten```/```nc-darken```-calls in ```variables.scss```, e.g. (not correct colors here, just for example):

```scss
$red: $color-error;				// warnings and errors
$red_overdue: #b3312d;		// overdue dates and high importance
$yellow: #fd0;						// medium importance
$blue_due: #4271a6;				// due dates and low importance
$blue_background: #168ad4;// highlighting date/timepicker dates
$gray_dark: nc-darken($color-main-background, 3%);
$gray: nc-darken($color-main-background, 10%);
$gray_light: nc-darken($color-main-background, 20%);
$gray_easy: nc-lighten($color-main-text, 60%);
$black: #000;
$white: #fff;
```

skjnldsv:
The problem is that variables are supposed to represent an entity not a color. Here you have various items like text colour, background, border...on a single variable! :)

skjnldsv:
I'm not against it, on the opposite, but right now the correct behavior isn't here! :)

raimund-schluessler:
I think I don't really understand your concerns. Do you want to have the variables named after their usage? Somehow like
```scss
$gray_light: nc-darken($color-main-background, 20%);
$gray_dark: nc-darken($color-main-background, 40%);

$border_color: $gray_light;
$background_color: $gray_light;
$text_color: $gray_dark;
```

I would be fine with this. But at the moment it makes things a lot harder, without any "real" reason, besides following a dogma 😉 .

skjnldsv:
@raimund-schluessler variables represent a specific utilisation, not just the color they have.
If I want to change all the borders colors of your app, I need to have this on a single variable. Right now, changing it will also change some background and text color that also use this variable. This needs to be split! :)

raimund-schluessler:
@skjnldsv I see your point and agree. But could you bring this branch in a usable state again, please? How the app currently looks like (see https://github.com/nextcloud/tasks/pull/108#issuecomment-331794252) it is hard to proceed with changes to the scss. Maybe we should even tackle the color issue in a follow-up PR.

skjnldsv:
I will revert my previous cleaning. We'll work on separate variables then! :)

raimund-schluessler:
@skjnldsv Mind having a look again? 😄

raimund-schluessler:
I think this is ready to be merged now. The category styling should be handled in a separate issue: https://github.com/nextcloud/tasks/issues/117

skjnldsv:
Sorry, please allow me two days to review. :)
Will do asap!

raimund-schluessler:
@skjnldsv I won't have internet for one week, so don't wonder when I won't answer. 😃

skjnldsv:
I agree with pixelipo :)
Changes looks good nonetheless!

raimund-schluessler:
@pixelipo @skjnldsv Thanks for having a look at this. 👍  I agree with all these points, I just think that some of them should better be solved in separate PRs.

>1. Shouldn't `css/src/sprite.scss` and `css/style.scss` be autogenerated (added to `.gitignore`)?
>2. Same question for `img/sprites.svg`, `js/public/app.min.js` and `js/public/app.min.js.map`

I did not do this here, because it wasn't like that before this PR and it's not related to the switch to SCSS in general. I used to have a working version of the app here in the repo, because I didn't want users having to compile the app. But with a lot of autogenerated files, I see that this becomes harder to maintain. Issue for removing the autogenerated files is in https://github.com/nextcloud/tasks/issues/118

>3. If it makes any difference, I'm strongly against chopping up SCSS into a dozen files. It helps in no way and only creates confusion about where is that particular declaration, or where a new one should be stored. I would argue that only `variables.scss` and perhaps 3rd party customizations (e.g. Jquery-ui code) deserve their own files.

I am fine with having only one SCSS file. I just introduced separate ones, because this way it was easier to cleanup and sort the different css rules. I will remove the separate files.

>4. Is it really a good idea to edit vendor files, such as `css/vendor/angularui/ui-select/select2.css`? I think those should be overridden instead.

It sure is not a good idea to edit vendor files and this should be cleaned. The issue is already in https://github.com/nextcloud/tasks/issues/117. But again, this is not strictly related to the move to SCSS, so I would like to handle this in a separate PR.

I will take care of the third point and would prefer to merge this PR here afterwards, because it is already quite big and the remaining remarks are not really about the switch to SCSS.

raimund-schluessler:
@pixelipo I removed the separate scss files and only use a single one now. I think this is good to merge.

raimund-schluessler:
Fix for the autogenerated files is in https://github.com/nextcloud/tasks/pull/119

skjnldsv:
Just a reminder, scss is only available for >12 :)

raimund-schluessler:
Thanks for the reviews! 

> Just a reminder, scss is only available for >12 :)

@skjnldsv This includes 12.0.0, right?

skjnldsv:
Yes, my bad, >=12! 🙈

raimund-schluessler:
@skjnldsv @pixelipo Thanks a lot for reviewing and helping.

PR #112, opened 2017-09-09
Respect new php controller structure

raimund-schluessler:


raimund-schluessler:
Merging as well as it is only renaming.

PR #120, opened 2017-10-15
Move to npm5

raimund-schluessler:
Move back to NPM since version 5 is available. This might make things easier since one must not install yarn anymore.


pixelipo:
Looks good, @raimund-schluessler 
I did notice two (perhaps unrelated) issues:

1. favicon did not generate
2. "Add list" button in the sidebar does not work (no `input` is revealed on click)

raimund-schluessler:
@pixelipo Hm, I cannot reproduce both of your issues.

1. The favicon is not generated, it should be located in `img` folder.
2. The input shows when clicking on the button and adding a list works fine.

Could you check if this also happens on the master branch?

raimund-schluessler:
We might have to use `npm5` instead of `npm` to really use npm in version 5 (`npm` on my system adresses npm in version 3). Can someone confirm this?

pixelipo:
@raimund-schluessler I also had npm v3. I've installed npm5 and didn't notice any difference. Issues mentioned before are persisting and I can't test the master because yarn is not working correctly any more :(

raimund-schluessler:
> Issue with the input field is solved with #130

Interesting, maybe this was an issue of NC13 then.

raimund-schluessler:
Let's merge this after #130 then.

PR #123, opened 2017-10-26
Fix icon opacities

raimund-schluessler:
This PR fixes the opacities of several icons when hovered or active.

For icons which are interactive, opacity changes to 0.7 now when hovered (e.g. adding a subtask or starring a task). Other icons which only indicate something but cannot be clicked have no hover effect (e.g. the icon indicating whether or not a task has a note).

Also the icon showing if completed subtasks should be shown (the eye) now has a opacity of 0.7 when the tasks are shown.

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -513,10 +513,6 @@ ol[dnd-list] {
 		}
 	}
 
-	.task-star .icon:hover {
-		opacity: .7;
-	}
-
 	.subtasks-container {
 		margin-left: 35px;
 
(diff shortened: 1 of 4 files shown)


pixelipo:
Great idea to use class to aggregate this type of icon!

Is there a reason why you're not using full opacity?

raimund-schluessler:
> Is there a reason why you're not using full opacity?

`0.7` just seemed to be the default from looking at the files app: https://github.com/nextcloud/server/blob/master/apps/files/css/files.scss#L584

PR #130, opened 2017-11-30
[WIP] Adjustments for NC13, fixes #129

### LINKED ISSUE #129: Adjustments for NC13
- [ ] The left sidebar is broken, the icon is to far left.
- [ ] The calendarname just disappears when selected in the sidebar.
- [ ] When you want to add a list and click on "add list" the entry disappears and nothing is shown. Same happens when you want to edit a calendar or show the calendarlink.
- [ ] The styling of the delete entry in the app-navigation-dropdown is broken, also the entry disappears after the countdown. 😕 
- [ ] The date- and timepicker styling needs to be fixed.

So far this is everything I found, but the list might need to be extended.

@skjnldsv Mind having a look at the sidebar issues?

raimund-schluessler:
- [x] The left sidebar is broken, the icon is to far left.
- [x] The sidebar icons need to be fixed. Do we really have to dismiss the svg sprite?
- [x] The calendarname just disappears when selected in the sidebar.
- [x] When you want to add a list and click on "add list" the entry disappears and nothing is shown. Same happens when you want to edit a calendar or show the calendarlink.
- [x] The styling of the delete entry in the app-navigation-dropdown is broken, also the entry disappears after the countdown. 😕 
- [x] The date- and timepicker styling needs to be fixed.
- [x] Timepicker has wrong lineheight (might be a core issue, @jancborchardt Could you have a look, please?)
- [x] The icons in the sortorder dropdown are messed up.
- [x]  The edit input and colorpicker overlaps other app-navigation-entries.


raimund-schluessler:
This really sucks. If we want to fit the sidebar to the css guidelines, we cannot use a svg sprite anymore. Or am I wrong?

jancborchardt:
Also cc other @nextcloud/designers - especially @skjnldsv regarding the sidebar and dropdown, and @pixelipo about the timepicker :)

jancborchardt:
Hmm, I'm not entirely sure - I think it might be possible to use an SVG sprite with modifications. Maybe @skjnldsv knows that quicker.

Did you use the standardized app-navigation before and this only broke with 13? Or did you not use those guidelines at all before?

raimund-schluessler:
I tried to stick with the guidelines, but I guess I didn't achieve this completly.

skjnldsv:
@raimund-schluessler you mean an inline svg? In the html? :)

raimund-schluessler:
@skjnldsv It would be six icons in total. Just increasing the space to 44px would probably fix the top, left and bottom visibility of the sprite, but you would still see the rest of the sprite on the right side. However, we can solve this with a `::before` pseudo-element with appropriate size. I will try that tomorrow.

pixelipo:
Icons should be 16*16. The rest of the way to 44px should be handled by padding in button itself

raimund-schluessler:
This should be good to go now. Please have a look.

I fixed the icon of the navigation entries with a pseudo element. It is not quite guideline conform I guess, but we can keep the icon sprite file.

raimund-schluessler:
This will break compatibility with NC12. I adjusted the min-version.

css/src/style.scss
```diff
@@ -1153,6 +1132,16 @@ ol[dnd-list] {
 	}
 }
 
+/**
+	* timepicker fixes
+	*/
+
+.ui-timepicker-table td a {
```
pixelipo: It seems to me that only `display: block` is needed here.

In fact, we should consider moving it to the core css.

css/src/style.scss
```diff
@@ -20,101 +20,75 @@ $white: #fff;
 	*/
```
pixelipo: I think `$blue_background` should also be deprecated now - it was primarily used for *pickers, now it's only used to highlight input fields (in a way that doesn't seem needed or useful).

css/src/style.scss
```diff
@@ -20,101 +20,75 @@ $white: #fff;
...
-				opacity: 1;
+				.app-navigation-entry-edit.name {
+					display: inline-block;
+					position: relative;
+				}
 			}
 
-			&.icon {
-				opacity: 1;
+			&.animate-up {
```
pixelipo: Doesn't seem to work for hidden elements (see the little dot?):
image
Here's another view:
screenshot from 2017-12-04 15-06-14


I'd suggest dropping the animation and just hiding this with `display:none`

css/src/style.scss
raimund-schluessler: Good idea, the border color should be handled by core by now.

css/src/style.scss
raimund-schluessler: I will check.

css/src/style.scss
raimund-schluessler: Would be good to have it in core, but until this is in core, we should keep the fix here.

raimund-schluessler: We need
```
.ui-timepicker-table td a {
   display: block;
   width: 14px;
}
```
here. The width is needed as well. With
```
padding: 8px 7px;
line-height: 12px;
```
(see https://github.com/nextcloud/server/blob/7d9e47543482acd430a18b482cd61bdd5eecc3e9/core/css/styles.scss#L1118-L1120) the element would not be quadratic and the blue circle would be an ellipse.

css/src/style.scss
raimund-schluessler: I fixed this. `display: none` was the solution.

css/src/style.scss
raimund-schluessler: Same goes for datepicker.

raimund-schluessler: Hm, datepicker seems to be more tricky. Setting the width to 12 px leads to problems for numbers with two digits. --> Reverting.

raimund-schluessler: Ok, datepicker highlight is now also circular.

css/src/style.scss
raimund-schluessler: The box shadow was used on the input fields in the detail view. It made look a little bit more "fancy", but I am ok with removing them. --> Solved.

raimund-schluessler:
@pixelipo I resolved all the issues you mentioned. Please have a look again. The date- and timepicker still look a bit different (different font-size, different spacing between the blue circles), but I think this is better solved in core.

pixelipo:
@raimund-schluessler commits looks solid, but I'd like to test them in a running app as well (tomorrow morning).

P.S. I'm testing this with #120 merged - works quite well :+1:

raimund-schluessler:
@pixelipo Thanks for having a look. Interesting to know about #120.

raimund-schluessler:
@pixelipo Thanks for having a look!

jancborchardt:
Thanks a lot folks! :) Sorry, I was on vacation last week and not on the computer mostly.

PR #140, opened 2018-01-19
Update pt_PT.js

jacmatias:
Better grammar.


jacmatias:
Better grammar in Portuguese Translation

rullzer:
Thanks for this! @jacmatias 

We maintain our translations on transifex. Please join there. The translations are then automatically synced: https://www.transifex.com/nextcloud/nextcloud/dashboard/

PR #142, opened 2018-02-06
Compatibility with NC12

raimund-schluessler:
This should fix all issues with NC12 so that master is compatible with both NC12 and NC13.

The file `css/src/nc12.scss` will be deleted once NC12 is EOL.

@skjnldsv Could you have a look and check that it works with NC12 and NC13, please?

### DIFF
--- a/css/src/nc12.scss
+++ b/css/src/nc12.scss
@@ -72,3 +72,191 @@
 	box-shadow: unset;
 	filter: drop-shadow(0 1px 3px rgba(179, 179, 179, .75));
 }
+
+$border-radius: 3px;
+
+/* ---- jQuery UI datepicker ---- */
+.ui-widget.ui-datepicker {
+	margin-top: 10px;
+	padding: 4px 8px;
+	width: auto;
+	border-radius: $border-radius;
+	border: none;
+	background: unset;
+	background-color: $color-main-background;
+	filter: drop-shadow(0 1px 3px rgba(179, 179, 179, .75));
+
+	.ui-state-default,
+	.ui-widget-content .ui-state-default,
+	.ui-widget-header .ui-state-default {
+		border: 1px solid transparent;
+		background: inherit;
+	}
+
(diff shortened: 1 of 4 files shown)


raimund-schluessler:
Ok, with this fix, the app is usable on NC12. There are some visual differences to the version on NC13, but only minor ones.

css/src/nc12.scss
```diff
@@ -72,3 +72,205 @@
 	box-shadow: unset;
 	filter: drop-shadow(0 1px 3px rgba(179, 179, 179, .75));
 }
+
+$color-main-text: #000;
```
skjnldsv: This is not needed, nc12 has these variables as well.
https://github.com/nextcloud/server/blob/stable12/core/css/variables.scss

css/src/nc12.scss
```diff
@@ -72,3 +72,205 @@
...
+
+		.ui-icon {
+			opacity: .5;
+
+			&.ui-icon-circle-triangle-e {
+				background: url('../../../apps/tasks/css/../img/arrow-right.svg') center center no-repeat;
+			}
+
+			&.ui-icon-circle-triangle-w {
+				background: url('../../../apps/tasks/css/../img/arrow-left.svg') center center no-repeat;
```
skjnldsv: This looks like a compiled path. use the correct url and let the scss compiler rewritter do its job :)

css/src/nc12.scss
raimund-schluessler: Thanks, I will remove it.

css/src/nc12.scss
raimund-schluessler: What would be the correct path? From the scss file to the svg file? Like `../img/arrow-left.svg`?

skjnldsv: Yes, relative to the task app directory :)

raimund-schluessler:
Fixed and rebased

PR #165, opened 2018-04-18
OpenTask-Support

newhinton:
I have opened a feature request regarding the OpenTasks-android-app, and i have started developing this feature. Since i have progressed quiet far, i wanted to open a pull request. There are still issues with my code, but i hope to get it fixed soon and or recieve help here.

[Feature Request](https://github.com/nextcloud/tasks/issues/159)

Current Issues: 
List elements are not updating their state after they were deleted (or checked). This only affects the UI, the changes are stored properly.

Adding Tasks is not the most elegant solution.

Hiding checklisttasks does not work, maybe this is related to the first issue.

If i have missed requirements for this pull request, please tell me what is missing, i try to get it fixed then.  Any help is appreciated.

PR #177, opened 2018-06-18
OpenTask-Support

raimund-schluessler:
Follow-up to #165


raimund-schluessler:
@newhinton What do I have to do in order to see such a checklist task? For me they don't show up for some reason.

js/app/controllers/settingscontroller.js
```diff
@@ -48,6 +48,15 @@ angular.module('Tasks').controller('SettingsController', [
 						name: t('tasks', 'Automatic')
 					}
 				];
+				this._$scope.parsingOptions = [
```
raimund-schluessler: I think the changes to this file are not used, since the value if the checklists tasks should be shown are stored per task.

css/src/style.scss
```diff
@@ -634,6 +647,133 @@ ol[dnd-list] {
 	}
 }
 
+.task-body-checklist {
```
raimund-schluessler: This should be merged with the .task-body rule as far as possible.

templates/part.taskbody.php
```diff
@@ -35,6 +35,12 @@
             title="<?php p($l->t('Toggle subtasks')); ?>">
         </span>
     </a>
+    <a class="handler" ng-click="toggleChecklist(task)">
```
raimund-schluessler: If this is per task, we should use a different icon than for the subtasks toggling.

raimund-schluessler:
I am not sure, if we should make the setting to show checklist tasks a per task setting. Maybe a global setting would be better?

raimund-schluessler:
Also @jancborchardt could you have a look if this feature is useful as it is now, screenshots are in #165.

newhinton:
I will take a peak why they are not showing up, but it will have to wait until the weekend, then i will find out why its not working anymore

jancborchardt:
@raimund-schluessler seems good to me, looks good also! :) I haven’t seen a similar feature anywhere yet, but as it will only show for people who use it, it seems great! :)

raimund-schluessler:
Any news here?

raimund-schluessler:
I want to, but it does not work yet. @newhinton Do you have time to look into this?

newhinton:
@raimund-schluessler are you at the conf?

raimund-schluessler:
> @raimund-schluessler are you at the conf?

Sadly not. I would have liked to, but I don't have much freetime at the moment.

newhinton:
@raimund-schluessler I was working on this on the weekend, but sadly i did not achieve any progress. The main problem at the moment is that the checkboxstate does not toggle its visual representation to reflect the change, and that the checklistelement is not able to be toggled.

I see that i have accidentially commited the changed package.json file, it incudes a watch script for auto-building. I am removing it in the next commit, or do we want to keep it?

pixelipo:
So you guys actually decided to go with this approach to show subtasks in the main window: https://github.com/nextcloud/tasks/pull/165#issuecomment-382477323

I'm still of the opinion that this is not a very clean approach, it mixes subtasks and checklist items in a not very user-friendly manner.

It might actually even degrade usability of the app for usecases where there are dozens (or more) checklist items.

I would love to hear more input from @nextcloud/designers 

my vote is to simply support checklist in the sidebar + make task completion percentage dependant on checklist

raimund-schluessler:
I am closing this, since the current state is non-functional and it needs to be reimplemented for the vue rewrite anyway. Also, the vue rewrite supports tasks lists in the notes field, so maybe this helps already.

PR #185, opened 2018-08-06
Layout fixes for NC14, closes #184

### LINKED ISSUE #184: Compatibility with Nextcloud 14
Any chance you can try to look into that @raimund-schluessler  ?

raimund-schluessler:
This fixes the layout for NC14 but will most likely break layout for NC<14.
@skjnldsv @nickvergessen 

PS: The diff for templates/main.php is huge, because the indentation changed (I removed the outer div).


nickvergessen:
https://github.com/nextcloud/tasks/pull/185/files?w=1 is here to help with the diff, it ignores whitespace changes

nickvergessen:
Didn't test, but from what I saw on other repos it should fix it

css/src/style.scss
```diff
@@ -327,7 +326,7 @@ $white: #fff;
 		margin: 12px 0;
```
skjnldsv: You ca use the new add button standard! :) 
https://github.com/nextcloud/server/blob/master/core/css/apps.scss#L99
```html 
<!-- app-navigation -->
<div class="app-navigation-new">
<button type="button" id="new-user-button" class="icon-add">New user</button>
</div>
<!-- ul -->
```

raimund-schluessler: You mean to use the new button to create new tasks?

skjnldsv: Yes. You can look the 14 documentation. On the design section. :)

PR #194, opened 2018-08-29
Move app to Vue.js

raimund-schluessler:
I started migrating the app to Vue.js now.

Basically nothing is functional at the moment, but the build is working and the store and the router are in place and operative. Furthermore, all templates are rudimentary moved to vue (the angular attributes are still present as a memory aid) and the collections are loaded from the server.

Any feedback is appreciated, especially on:
- how to load the tasks from the server (shared caldav library?)
- how to keep the tasks in the store


skjnldsv:
Quick look and feedbacks :)
- you should use the same makefile functiosn names than the other nc app for clarification accross all of our apps. Easier for people to understand ;)
- Please also ouse the same directory structure than the other. You can find a vue template [here](https://github.com/nextcloud/vue-components/tree/master/packages/webpack4-template-simple) and a vue nc app example [here](https://github.com/skjnldsv/vueexample) (very rough one we used as a sandbox, but maybe you will find this interesting :)

Questions:
- why still use gulp, webpack should be enough? I'm not the most expert in gulp, so I have no idea of the advantages of having both :)

raimund-schluessler:
Thanks for the feedback. 😃 

> you should use the same makefile functiosn names than the other nc app for clarification accross all of our apps. Easier for people to understand ;)

I will have a look.

> Please also ouse the same directory structure than the other.

Means, the vue/js files should be in `src`?

> why still use gulp, webpack should be enough?

I would like to get rid of it as well. It currently is used to compile the sprite file and concat the scss files. I will have a look how to do this with webpack only.

skjnldsv:
> Means, the vue/js files should be in src?

Yes :)

> I would like to get rid of it as well. It currently is used to compile the sprite file and concat the scss files. I will have a look how to do this with webpack only.

Well, you can create a single tasks.scss app which import all the other scss, the server will then compile a single file ;)
For the sprite file, I have no idea :(

skjnldsv:
> I couldn't get rid of gulp completely yet, but it might be possible by using a plain npm module. I still have to adjust the makefile, though.

Maybe it's just required for your setup then :)
It's not necessarily a pas thing ;)

src/components/TheList.vue
```diff
@@ -171,6 +170,19 @@ export default {
 			calendars: state => state.calendars,
 			dayOfMonth: state => state.dayOfMonth
```
skjnldsv: Nice, I forgot about this way of using vuex!! :+1:

src/app.vue
```diff
@@ -0,0 +1,143 @@
...
+		</div>
+
+		<div id="app-sidebar">
+			<theDetails />
+		</div>
+	</div>
+</template>
+
+<script>
+import TheList from './components/TheList'
```
skjnldsv: We already have a component for the Navigation :)
https://github.com/nextcloud/vue-components/tree/init

src/components/PopoverMenu.vue
```diff
@@ -0,0 +1,62 @@
+<!--
+Nextcloud - Tasks
```
skjnldsv: We also already have a component for the popoverMenu which is standardised :)

src/app.vue
raimund-schluessler: Thanks for the hint. Do you use them in the contacts rewrite already? I can't find how you imported them there?

skjnldsv: Yes, we copied them directly because we did not published them to npm yet. But this should be done next week :)
Sorry about the late comment, i thought I already give you this information :(

You can also see that we imported them into the vueexample app (see package.json). If you use the same components from this repo the transition to import should be very straightforward anyway ;)

raimund-schluessler: I guess a
`
"vue-components": "git+https://github.com/nextcloud/vue-components.git#init",
`
should do for now. I will have a look how they work.

raimund-schluessler: A problem with the popover component will be, that I cannot use a custom item there. The delete item is currently showing an element asking if you really want to delete: https://github.com/nextcloud/tasks/blob/master/js/app/directives/confirmation.js. That will not work with the standard popover menu.

skjnldsv: You should not use any custom item according to our standard though :)
That's why we created standards :see_no_evil: 

The timeout is not necessary if you already have an undo option.

skjnldsv: And if you want a confirmation, just change the object item text and ask for a second click action, still within the same object :)

skjnldsv: Arf, sorry, I thought you had an undo function! My bad! :)

Well, the components use a dedicated prop which is an object.
After clicking the delete action, you can just trigger a change that replace the object entry related to the delete action by a confirm one that will then trigger the final deletion :)

You can even have the countdown if you want, when using a computed variable, it should be feasible!

Exampel of dynamic checks to build the menu: https://github.com/nextcloud/contacts/blob/vue/src/components/Settings/SettingsAddressbook.vue#L129
Or dynamic text: 
https://github.com/nextcloud/contacts/blob/vue/src/components/Settings/SettingsAddressbook.vue#L102

skjnldsv: Rough example
``` js
			menu = [
				{
					icon: 'icon-public',
					text: this.confirm
						? t('tasks', 'Are you sure')
						: t('tasks', 'Delete'),
					action: this.confirm
						? this.delete
						: this.toggleConfirm
				}
			]
			return menu
 ```

Comment (raimund-schluessler) [reply src/app.vue]: Thanks a lot, I will have a look. 😃

File: src/components/Confirmation.vue [hunk 4]
```diff
@@ -0,0 +1,90 @@
...
+		}
+	},
+	methods: {
+		reset: function() {
+			this.activated = false
+			this.armed = false
+			this.remaining = 3
+		},
+		activate: function(e) {
+			$('.confirmation-confirm').tooltip({ title: this.message, container: 'body', placement: 'right' })
```
skjnldsv: Also, please, do not use jQuery as we want to get rid of it on the following releases :)

skjnldsv: https://github.com/nextcloud/vue-components/issues/9#issue-351801788
Tooltip replacement use in nextcloud: https://github.com/Akryum/v-tooltip

raimund-schluessler: I removed the first usage of jQuery and tooltip in 35cfab2f949c85d3ddece78416be44b35d623d98. Will remove the rest soon.

raimund-schluessler: No jQuery anymore.

src/components/TheCollections/Week.vue
```diff
@@ -20,47 +20,71 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
...
+			var date = moment().add(day, 'day')
+			var dayString
+			if (day == 0) {
+				dayString = t('tasks', 'Today')
+			} else if (day == 1) {
+				dayString = t('tasks', 'Tomorrow')
+			} else {
+				dayString = date.format('dddd')
+			}
+			return dayString + ', ' + date.format('LL')
```
raimund-schluessler: @skjnldsv @georgehrke Do you know, if I have to configure the locale manually for the app? When I select a locale under settings->user and navigate to the Tasks app, it is shown using the correct locale. But when I reload the page, the locale is lost and the english fallback is used. However, under settings, the correct locale is still shown. Should moment.js be configured with the correct locale by the server automatically?

skjnldsv: @raimund-schluessler Take a look at my implementation of the dateTime with dynamic locale load and moment
https://github.com/nextcloud/contacts/blob/vue/src/components/Properties/PropertyDateTime.vue

:)

raimund-schluessler: Thanks. Is there a reason why you don't use the moment.js version built into server?

skjnldsv: @raimund-schluessler I can manage the updates myself and not depends on the server :)

raimund-schluessler: Also, this doesn't seem to be necessary for Calendar?!

skjnldsv: @raimund-schluessler what is not nexessary? :)

Also, reference of config and nice setup to not load all the locale at once: https://github.com/nextcloud/vue-components/issues/13#issuecomment-419452405

raimund-schluessler: > @raimund-schluessler what is not nexessary? :)

To configure the locale manually. At least I couldn't find it anywhere in the Calendar app code.

skjnldsv: Ah, no idea. :thinking:

raimund-schluessler: This https://github.com/nextcloud/tasks/pull/194/commits/9c6ca2b241d95932a1161eab520ad28ef9ed8559 seems to do the trick for now. :)

webpack.common.js
```diff
@@ -44,9 +44,6 @@ module.exports = {
 	},
 	plugins: [new VueLoaderPlugin(), new StyleLintPlugin()],
 	resolve: {
-		alias: {
```
raimund-schluessler: @skjnldsv This might also be an option for the contacts app. See https://blog.wuc.me/2018/10/18/nextcloud-vue-csp.html and https://vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only. For Tasks, this reduced the dev build by ~200 kB.

skjnldsv: @raimund-schluessler Didn't we already do that? :O :D

src/store/calendars.js
```diff
@@ -336,7 +338,10 @@ const actions = {
 					return mapDavCollectionToCalendar(calendar)
 				})
 			})
-
+		
+		// remove calendars which don't support tasks
+		calendars = calendars.filter(calendar => calendar.supportsTasks)
```
raimund-schluessler: @skjnldsv @georgehrke Is this a good solution to filter VTODO calendars or should this be in the CDAV library?

src/store/calendars.js
```diff
@@ -123,12 +172,379 @@ const getters = {
...
+
+	/**
+	 * Append a new calendar to array of existing calendars
+	 *
+	 * @param {Object} context the store mutations
+	 * @param {Object} calendar The calendar to append
+	 * @returns {Promise}
+	 */
+	async appendCalendar(context, calendar) {
+		return client.calendarHomes[0].createCalendarCollection(calendar.displayName)
```
raimund-schluessler: @skjnldsv @georgehrke Is there a way to specify the calendar color on creation? Currently no calendar color is given which results in a transparent calendar icon. You can change the color later, though.

src/store/calendars.js
skjnldsv: @raimund-schluessler filter can be crazy expensive on ressources.
I'll suggest you make performances tests to ensure this is not a bottleneck for your app :)

src/store/calendars.js
skjnldsv: @raimund-schluessler there should be a way to send the proper keys on create I guess, but you need to see with @georgehrke for that ;)

georgehrke: Color is the second parameter :)
https://github.com/nextcloud/cdav-library/blob/master/src/models/calendarHome.js#L108

I will also add a third parameter to define supported calendar component sets, so that you can create "pure" VTodo calendars.

raimund-schluessler: > I will also add a third parameter to define supported calendar component sets, so that you can create "pure" VTodo calendars.

Perfect.

src/store/calendars.js
raimund-schluessler: Well, this should only be called once on app load. But I am open to other solutions ;)

skjnldsv: Looking at the code, the objects are quite simple, so it should be fine, can you do a performance tests before and after the filter to check?

raimund-schluessler: I don't have much experience with performance testing. Should I just do a `console.time('Filter') ... console.timeEnd('Filter')` or what do you recommend?

src/store/calendars.js
```diff
@@ -0,0 +1,580 @@
...
+	/**
+	 * Retrieve the tasks of the specified calendar
+	 * and commit the results
+	 *
+	 * @param {Object} context the store mutations
+	 * @param {Object} importDetails = { ics, calendar }
+	 * @returns {Promise}
+	 */
+	async getTasksFromCalendar(context, { calendar }) {
+		return calendar.dav.findByType('VTODO')
```
raimund-schluessler: I get a bad request answer from the server for this line.

Parameters:
`<x1:calendar-query xmlns:x1="urn:ietf:params:xml:ns:caldav"><x0:prop xmlns:x0="DAV:"><x0:getcontenttype/><x0:getetag/><x0:resourcetype/><x0:displayname/><x0:owner/><x0:resourcetype/><x0:sync-token/><x0:current-user-privilege-set/><x0:getcontenttype/><x0:getetag/><x0:resourcetype/><x1:calendar-data/></x0:prop><x1:comp-filter name="VCALENDAR"><x1:comp-filter name="VTODO"/></x1:comp-filter></x1:calendar-query>`

Answer:
`<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
  <s:message>The {urn:ietf:params:xml:ns:caldav}filter element is required for this request</s:message>
</d:error>`

@georgehrke Do you have an idea what might be wrong?

src/store/calendars.js
raimund-schluessler: Takes 1ms for five calendars: `calendars.filter: 1ms`
This seems to be ok for the initial loading time. I will check the other places where I use `filter` as well.

skjnldsv: Ahaha, perfect :p

src/store/calendars.js
raimund-schluessler: I guess something like `<x1:filter> ... </x1:filter>` surrounding the `<x1:comp-filter name="VCALENDAR"><x1:comp-filter name="VTODO"/></x1:comp-filter>` might be missing?

georgehrke: Possibly, will look into it tomorrow

raimund-schluessler:
@skjnldsv @georgehrke I implemented updating/changing tasks (so far only toggling the completed state, https://github.com/nextcloud/tasks/pull/194/commits/574db7200ce4a23485a55cd5ee80aa05a70ead2b) and found a weird behaviour. For newly created tasks, the second and all following changes after task creation are failing with an ETag mismatch.

I don't really understand why this is happening yet, but it might have something to do with the cdav-library. I tried to do some debugging, this is what I found out yet and where I am stuck:

- When I create a task, the server answers with a response header without an ETag set. In difference, using the Contacts app and creating a new contact the following response header contains an ETag.
- When doing the next update, the If-Match property contains `null`, which leads the request to fail.
So, I guess the question is, why the response does not contain an ETag. Could this have something to do with the `Content-Type`? The content type is different when doing an `dav.createVObject` and `dav.update`.

Any help would be appreciated.

raimund-schluessler:
There also is a second problem. When a second `dav.update` is send, before the first is finished, the second will fail, because the ETag has changed in between. I understand why it is happening, but I would like to hear how you solve this in the Contacts or Calendar app. @skjnldsv @georgehrke.

skjnldsv:
> There also is a second problem. When a second `dav.update` is send, before the first is finished, the second will fail, because the ETag has changed in between. I understand why it is happening, but I would like to hear how you solve this in the Contacts or Calendar app. @skjnldsv @georgehrke.

We use a debounce function for every update.
Nonetheless, if a request takes forever, and a second one takes a few ms, I think it might still be an issue.  
Though I played a bit with the contacts app and coud not break the process. It works every time.

On a side note, we now have a conflict system on contacts. If you reach a 412, you have a warning and you can manually refresh the contact and discard your changes. This is a nice UX since you have an explanation on why it failed and what is going on.

src/components/Colorpicker.vue
```diff
@@ -0,0 +1,179 @@
...
+			this.pick(this.random)
+		},
+		pick: function(color) {
+			this.$emit('color-selected', color)
+		},
+		/*
+		* Generate a random colour with the core generator
+		*/
+		randColour: function() {
+			if (typeof String.prototype.toHsl === 'function') {
```
skjnldsv: We change this a while ago :)
https://github.com/nextcloud/server/blob/master/core/js/placeholder.js#L64

raimund-schluessler:
I tend to merge this into master now. We have basically reached feature parity with master, besides the missing implementation of drag and drop. Since, dnd is broken in master (for chrome at least) anyway, it might not be such a huge regression.

I would like to merge it, so that the translations are being updated in time, because there were quite a few changes. Also, we can close a lot of issues then 😉 

I will implement dnd before a release and fix bugs we find.

@skjnldsv @jancborchardt @georgehrke @pixelipo What do you think? And are you up to a review? 😃

skjnldsv:
Congrats!!! :tada:

nachoparker:
Many thanks for this! can't wait to try it out

dmp1ce:
Wow! I just received v0.10.0 and I can now use my tasks tab again. Thank you and nice work!

Orhideous:
Works like a charm. Thank you!

nachoparker:
thank you, it works very well. Just two little points that I'll address in other issues but performance is so much better

michalroxorpl:
Hello everyone, this modification is absolute awesome :-)
I only wish, there would be a better support for subtasks / markdown understanding in terms of how new subtasks are recognized.
The nextcloud tasks addon requires following format of subtasks:
`- [ ] description`
However, the OpenTasks android application supported by nextcloud / dav sync uses following format:
`[ ] description`

This small difference makes what is added in an OpenTasks app as a subtask visible as text on the nextcloud tasks subpage, on the other hand inserting a Markdown formatted subtask / internal task on nextcloud tasks subpage produces a non-formatted string with square brackets, that is not understood as subtask and not rendered correctly by OpenTasks.

raimund-schluessler:
@michalroxorpl Could you please create a separate issue for this? Maybe we can adjust the configuration of the Markdown plugin we use to also recognize tasks without the leading dash.

raimund-schluessler:
@michalroxorpl I just had a quick look at the markdown-it documentation and it seems a list item has to start with `-`, `+` or `*` (and tasks are list items as well).

So while all of this
```
- [ ] Test0
+ [ ] Test1
* [ ] Test2
```

will be a task item, omitting the leading `-`, `+` or `*` is not following the markdown standard. So it would be best to adjust the behaviour of OpenTasks. Please file an issue there.

PR #223, opened 2018-12-12
Tasks sorting.

nicolad:
@raimund-schluessler, this is a raw version. I am not sure if passing `sortOrder, sortDirection` along with tasks is a good option, just wanted to have something from which I can iterate over and improve.
I think sorting should be closer to tasks store, rather than being coupled with one of collection views.
What do you think ?

### DIFF
--- a/src/components/Task.vue
+++ b/src/components/Task.vue
@@ -3,6 +3,7 @@ Nextcloud - Tasks
 
 @author Raimund Schlüßler
 @copyright 2018 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+@copyright 2018 Vadim Nicolai <contact@vadimnicolai.com>
 
 This library is free software; you can redistribute it and/or
 modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
@@ -107,7 +108,7 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 				</form>
 			</div>
 			<ol v-if="!task.hideSubtasks" :calendarID="task.calendar.uri">
-				<TaskBodyComponent v-for="subtask in task.subTasks"
+				<TaskBodyComponent v-for="subtask in sort([...task.subTasks], sortOrder, sortDirection)"
 					:key="subtask.uid"
 					:task="subtask"
 					class="subtask"
@@ -119,9 +120,9 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 </template>
 
 <script>
-import { overdue, valid } from '../store/storeHelper'
+import { overdue, valid, sort } from '../store/storeHelper'
 import ClickOutside from 'vue-click-outside'
-import { mapActions } from 'vuex'
+import { mapGetters, mapActions } from 'vuex'
 import focus from '../directives/focus'
 
 export default {
@@ -158,6 +159,10 @@ export default {
 		}
 	},
 	computed: {
+		...mapGetters({
+			sortOrder: 'sortOrder',
+			sortDirection: 'sortDirection'
+		}),
 		/**
 		 * Returns the path of the task
 		 *
@@ -207,6 +212,7 @@ export default {
 			'toggleCompleted',
 			'toggleStarred'
 		]),
+		sort,
 		/**
 		 * Checks if a date is overdue
 		 */
--- a/package-lock.json
+++ b/package-lock.json
@@ -6142,8 +6142,7 @@
 				"inherits": {
 					"version": "2.0.3",
 					"bundled": true,
-					"dev": true,
-					"optional": true
+					"dev": true
 				},
 				"ini": {
 					"version": "1.3.5",
--- a/src/components/TheCollections/Calendar.vue
+++ b/src/components/TheCollections/Calendar.vue
@@ -3,6 +3,7 @@ Nextcloud - Tasks
 
 @author Raimund Schlüßler
 @copyright 2018 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+@copyright 2018 Vadim Nicolai <contact@vadimnicolai.com>
 
 This library is free software; you can redistribute it and/or
 modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
@@ -43,7 +44,7 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 					collectionId="uncompleted"
 					type="list"
 				>
-					<Task v-for="task in uncompletedRootTasks(calendar.tasks)"
+					<Task v-for="task in sort(uncompletedRootTasks(calendar.tasks), sortOrder, sortDirection)"
 						:key="task.id"
 						:task="task"
 					/>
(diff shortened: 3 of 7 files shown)


raimund-schluessler:
I am not 100% decided where to put the sorting function.

But since, we also need to sort the subtasks, I tend to not have the sorting function inside the store. The subtasks come directly from the parent task (see https://github.com/nextcloud/tasks/blob/vue/src/components/Task.vue#L110), so wee need access to the sort function in the Task component as well.

Maybe importing the sort function into the components would be an option. It could be done similar to https://github.com/nextcloud/tasks/blob/vue/src/components/Task.vue#L122 and then be applied to the root tasks from the store and the subtasks from the parent task.

raimund-schluessler:
Looks great so far! Just two things:

- It might be better to put the `reverse` into `sort` instead of `sortAlphabetically` because it is needed for the other sortOrders as well.
- Can I ask you to put your copyright into the files you edited and sign off your commits? See 
https://github.com/nextcloud/server/blob/master/contribute/HowToApplyALicense.m and 
https://github.com/nextcloud/server/blob/master/.github/CONTRIBUTING.md#sign-your-work for details.

nicolad:
Have done those adjustments.
I wanted to ask how should be treated tasks with empty `due` or `start` property on sorting - should those be placed at the begging of the array or at the end ?

src/components/Task.vue
```diff
@@ -158,6 +159,10 @@ export default {
 		}
 	},
 	computed: {
+		...mapGetters({
```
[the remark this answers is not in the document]
raimund-schluessler: I wanted to use it as well, but haven't had a look yet how to enable it, thanks for taking care. So 👍

raimund-schluessler:
> I wanted to ask how should be treated tasks with empty due or start property on sorting - should those be placed at the begging of the array or at the end ?

The reference is the old Angular.js version. There, they are shown at the end.

src/components/Task.vue
```diff
@@ -107,7 +108,7 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 				</form>
 			</div>
 			<ol v-if="!task.hideSubtasks" :calendarID="task.calendar.uri">
-				<TaskBodyComponent v-for="subtask in task.subTasks"
+				<TaskBodyComponent v-for="subtask in sort(task.subTasks, sortOrder, sortDirection)"
```
raimund-schluessler: For some reason this line gives me an error:
warn

nicolad: I made this change https://github.com/nextcloud/tasks/pull/223/commits/04b1bdeaed232a66a8cd4d864901e01e778caa12
according to Evan's comment: https://github.com/vuejs/vue/issues/3276#issuecomment-232818665

raimund-schluessler: Interesting, I didn’t know that.

raimund-schluessler:
> What do you think ?

Couldn't we adjust the sort function to account for that already?

nicolad:
> > What do you think ?
> 
> Couldn't we adjust the sort function to account for that already?

Yes, we can: https://github.com/nextcloud/tasks/pull/223/commits/416903e89002e1388e16648125433f8c620ae0e7

raimund-schluessler:
Btw this is how I solved it in the angular branch back then 
https://github.com/nextcloud/tasks/blob/master/js/app/controllers/taskscontroller.js#L334

nicolad:
> Btw this is how I solved it in the angular branch back then
> https://github.com/nextcloud/tasks/blob/master/js/app/controllers/taskscontroller.js#L334

Thx, adjusted as it was there:
https://github.com/nextcloud/tasks/pull/223/commits/cb3fd7348e191bc8bbcea72f43df15bf37118d10

raimund-schluessler:
@MorrisJobke Could you please add @nicolad to the Nextcloud organization, so he has access to the Tasks repo?

raimund-schluessler:
@jancborchardt Maybe you could add @nicolad to Nextcloud=?

jancborchardt:
Added @nicolad to @nextcloud/tasks! Welcome to Nextcloud! :tada: 

Also @nicolad could you add screenshots so it’s easier to review from a design side? :) Thanks!

raimund-schluessler:
> Added @nicolad to @nextcloud/tasks! Welcome to Nextcloud! 🎉

Thanks a lot!

> Also @nicolad could you add screenshots so it’s easier to review from a design side? :) Thanks!

In principle screenshots are good, but here we work on the vue branch and only implement the functionality of the master branch again using vue.js. So no design changes here. 😉

nicolad:
Thank you guys :)
I am glad that I am able to contribute to Nextcloud

raimund-schluessler:
@nicolad I had a look at the PR now. There are a few things:

- https://github.com/nextcloud/tasks/commit/cb3fd7348e191bc8bbcea72f43df15bf37118d10 does not work properly. It gives ```[Vue warn]: Error in render: "Error: invalid date-time value: "last""``` I think splitting the array by empty dates and merging it again gives quite some overhead. It might be better if the sort function https://github.com/nextcloud/tasks/pull/223/files#diff-739e67d6112be1fb3f1b500f9dfe526cR176 would check if a date is null and return the approriate value (-1,0,1) in that case.
- In the calendar views only the completed tasks are sorted. You missed adding the sort to https://github.com/nextcloud/tasks/blob/cb3fd7348e191bc8bbcea72f43df15bf37118d10/src/components/TheCollections/Calendar.vue#L47

It would be good, if you could fix the two problems and then squash all commits into a few. I would merge it then.

There are a few more things, but we can solve them with a follow-up PR:

- I think the sort direction is reversed compared to the master branch (at least for the alphabetical sorting). I need to check, fixing would be easy.
- The default sort order is not implemented yet. Tasks were sorted by the properties `completed`, `dueDate`, `priority`, `startDate` and `summary` see https://github.com/nextcloud/tasks/blob/master/js/app/controllers/taskscontroller.js#L363. This one would be more difficult, because sorting by multiple properties would be necessary. I would like to add this as well, because for subtasks, there should be the possibility to show completed subtasks last, but still sort by due-date. However, we can add this later.

nicolad:
@raimund-schluessler I have adjusted those 2 issues. Also had squashed commits and rebased upstream vue branch.

Adding additional changes in a follow-up PR is a good idea, since we will have another thread of discussions there and will keep this one shorter :)

raimund-schluessler:
@nicolad Thanks a lot, but there are two new errors now.

The rebase of https://github.com/nextcloud/tasks/pull/223/files#diff-45ed2c4797f9a410abd706c99fcdd28c failed, there are syntax errors now and duplicated methods.

Also, this line https://github.com/nextcloud/tasks/pull/223/files#diff-739e67d6112be1fb3f1b500f9dfe526cR175 should return 1.

Once this is fixed, we will merge.

nicolad:
Done, Herr @raimund-schluessler

raimund-schluessler:
Very nice, thanks a lot for taking care! Expecting more PRs from you 😉

PR #225, opened 2019-01-06
Linkify links in task title

raimund-schluessler:
This would linkify links in the summary of a task.

But the dependency to https://github.com/punkave/sanitize-html bloats the package size by ~250 kB. I think this is to much for such a simple feature. I will look into a different solution.

@skjnldsv @juliushaertl @georgehrke Any ideas what to use maybe?

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -356,7 +356,7 @@
 			overflow-y: auto;
 		}
 
-		.task-body .title-wrapper span.title a {
+		.task-body .task-info-wrapper .title span a {
 			cursor: pointer;
 			text-decoration: underline;
 		}
--- a/package-lock.json
+++ b/package-lock.json
@@ -6448,14 +6448,12 @@
 				"balanced-match": {
 					"version": "1.0.0",
 					"bundled": true,
-					"dev": true,
-					"optional": true
+					"dev": true
 				},
 				"brace-expansion": {
 					"version": "1.1.11",
 					"bundled": true,
 					"dev": true,
-					"optional": true,
 					"requires": {
 						"balanced-match": "^1.0.0",
 						"concat-map": "0.0.1"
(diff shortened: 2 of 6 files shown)


georgehrke:
I was looking at `linkifyjs` for calendar as well, but having a ~250kB dependency is a real bummer.

raimund-schluessler:
> I was looking at ```linkifyjs``` for calendar as well, but having a ~250kB dependency is a real bummer.

```linkifyjs``` itself is fine, I think. But we need to use it with ```v-html``` and this does not sanitize the HTML you hand over. Since this is user content, we need to sanitize it ourself. And the ```sanitize-html``` depends on ```postcss``` which bloats the size: https://github.com/punkave/sanitize-html/issues/198

I guess, either finding a better (smaller) sanitizer or a linkify library which does not require using ```v-html``` would help.

georgehrke:
Did you check https://www.npmjs.com/package/vue-linkify? 

It does not depend on sanitize-html, we should just carefully test it for XSS, etc.

raimund-schluessler:
> Did you check https://www.npmjs.com/package/vue-linkify?

I had a look, but didn't test the package. It uses ```el.innerHTML``` and does no sanitizing: 
https://github.com/phanan/vue-linkify/blob/master/index.js#L6. I suspect, it will be vulnerable too. There also is an issue discussing this: https://github.com/phanan/vue-linkify/issues/1. Plus, the package is two years old and uses an old version of linkifyjs. --> I would not use this.


Seems like we have to use ```linkify-string``` https://soapbox.github.io/linkifyjs/docs/linkify-string.html and manually insert it with ```v-html```. This should prevent XSS then: https://soapbox.github.io/linkifyjs/docs/caveats.html#cross-site-scripting.

The other option would be to use ```vue-markdown``` https://www.npmjs.com/package/vue-markdown which I want to use for the notes field anyway. But let's keep it simple for now.

raimund-schluessler:
I wrote a small directive https://github.com/nextcloud/tasks/pull/225/files#diff-815623279d244cd1963d6c0d7aee8f96 which uses linkify-string. I shouldn't be vulnerable as linkify-string escapes all html entities and I couldn't trigger any XSS problem. Tests are welcome though.

raimund-schluessler:
This increases the package size by 20 kB, which is ok, I guess.

PR #228, opened 2019-01-08
Vue markdown

raimund-schluessler:
This implements a markdown preview for the notes field. But it doubles the size of the package. 😞


raimund-schluessler:
Using `markdown-it` and plugins directly leads to ~150 kB package increase. I suppose this is the best we can get.

PR #229, opened 2019-01-20
Indicate the status of the sync process inline

raimund-schluessler:
This PR adds an icon indicating the status of the sync process for every task. Shows a spinner while the request is running, a green checkmark for 5 seconds when the request was successful and a refresh icon when an etag mismatch occured (can be clicked to fetch the task from the server).

This can be enhanced when further states are necessary (such was error, warning or info).

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -665,12 +665,17 @@
 				}
 			}
 
+			.status-display {
+				float: right;
+				margin: 17px 0;
+			}
+
 			.title-text {
 				cursor: text;
 				display: inline-block;
 				min-height: 24px;
 				margin: 13px 0;
-				width: calc(100% - 115px);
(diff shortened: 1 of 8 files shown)


raimund-schluessler:
Looks like this in task item:

Success:
screenshot_2019-01-20 aufgaben - nextcloud 1

Request running:
screenshot_2019-01-20 aufgaben - nextcloud 2

Etag mismatch:
screenshot_2019-01-20 aufgaben - nextcloud

And in details view:

Success:
screenshot_2019-01-20 aufgaben - nextcloud 4

Request running:
screenshot_2019-01-20 aufgaben - nextcloud 5

Etag mismatch:
screenshot_2019-01-20 aufgaben - nextcloud 3

PR #236, opened 2019-01-27
Implement drag and drop

raimund-schluessler:
This PR aims to implement drag and drop in order to

- [x] Move subtasks
- [x] Move tasks between calendars
- [x] Star a task, set the date to today or complete a task by dropping on corresponding collection

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -451,6 +451,14 @@
 		}
 	}
 
+	&.sortable-ghost .task-body {
+		background-color: rgba( $color-primary, .3 );
+	}
+
+	&.dragover > div.subtasks-container > ol {
+		min-height: 37px;
+	}
+
 	.subtasks-container {
 		margin-left: 35px;
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -13482,6 +13482,11 @@
         "kind-of": "^3.2.0"
       }
     },
+    "sortablejs": {
+      "version": "1.8.4",
+      "resolved": "https://registry.npmjs.org/sortablejs/-/sortablejs-1.8.4.tgz",
+      "integrity": "sha512-Brqnzelu1AhFuc0Fn3N/qFex1tlIiuQIUsfu2J8luJ4cRgXYkWrByxa+y5mWEBlj8A0YoABukflIJwvHyrwJ6Q=="
+    },
     "source-list-map": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/source-list-map/-/source-list-map-2.0.1.tgz",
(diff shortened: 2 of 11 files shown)


raimund-schluessler:
@nextcloud/vue Reviews and testing is welcome for this PR now. It's not entirely done yet, but the most important part with drag and drop for moving between calendars and subtasks should work.

I don't know whether vuedraggable/sortable.js is the best dnd library for Tasks, but it works for now. One thing which might be not so good, is that you cannot cancel a drag. If you move out of the drop zone the task will be inserted at the last valid position. What do you think?

ChristophWurst:
I also learned that dnd isn't that easy with Vue, or at least the libs that were available when I tried couldn't help implementing the message drag-and-drop in Mail.

skjnldsv:
Nice work!!
Sotty I'm late to the party!

Just a bit janky sometimes 
Capture d’écran_2019-05-06_15-28-52

raimund-schluessler:
> Sotty I'm late to the party!

@skjnldsv Sorry, I thought you had a look already.

> Just a bit janky sometimes

What do you mean with janky? I know making a task a subtask can be a bit annoying sometimes because you have to hover over the parent task and then move a bit down. It's the best I could come up with as nested items are rarely supported by dnd libraries (and it was like that in v0.9.8 already ;).

But we sure can check whether to change the dnd library for later versions.

skjnldsv:
I mean, it feels a tad weird moving tasks around, not smooth :)
When moving there is this weird empty gap that appear (in my screenshot, there is 5 items spaces, with the empty one, but it should be still 4? And have the empty space where the item is about to be dropped, no?)

Also, on another note, please make the default height to 44px like we ask our devs to do. :)
It's easier for everyone on mobile to properly click things (buttons are 44x44) :wink: 

> @skjnldsv Sorry, I thought you had a look already.

Don't be!! I'm the one who's late!! :hugs:

raimund-schluessler:
> I mean, it feels a tad weird moving tasks around, not smooth :)
When moving there is this weird empty gap that appear (in my screenshot, there is 5 items spaces, with the empty one, but it should be still 4? And have the empty space where the item is about to be dropped, no?)

Yes, this is due to the dnd library I think. I will look into finding a better dnd library for the next version. For now, I think it is best to use what we have, as it is better than the version before (works in Chrome now). And we need a release for NC16 😉 

> Also, on another note, please make the default height to 44px like we ask our devs to do. :)
It's easier for everyone on mobile to properly click things (buttons are 44x44) 😉

Which height do you mean? The height of a task? Could you please open a separate issue for this?

PR #401, opened 2019-05-11
Improve compatibility with weird browsers

raimund-schluessler:
e.g. Edge and Pale Moon


raimund-schluessler:
I will merge this for now. Further changes for Pale Moon compatibility would require a lot of effort and we lack the resources to support any browser. PRs are welcome though.

PR #410, opened 2019-05-16
Implement sharing calendars / task lists

### LINKED ISSUE #14: [Feature Request] Share task lists
Hi, 

will there be any futhur plan to add a share function, to share task lists with other users?

raimund-schluessler:
Implements a sharing menu in the calendar list. You can now share task lists with other users. Closes #14.

@skjnldsv I shamelessly copied most of the code from the Contacts app 😇 Works very good as far as I can tell.

Reviews very welcome!

### DIFF
--- a/css/src/Calendars/Calendar.scss
+++ b/css/src/Calendars/Calendar.scss
@@ -1,0 +1,64 @@
+/**
+ * @copyright Copyright (c) 2019 John Molakvoæ <skjnldsv@protonmail.com>
+ *
+ * @author John Molakvoæ <skjnldsv@protonmail.com>
+ * @author Team Popcorn <teampopcornberlin@gmail.com>
+ * @author Raimund Schlüßler <raimund.schluessler@mailbox.org>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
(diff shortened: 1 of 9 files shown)


raimund-schluessler:
I will merge this now, so that the translations are getting updated. Reviews and testing is still welcome. If issues are found, we can fix them in later PRs.

marbetschar:
@raimund-schluessler not quite sure where to put this, so I shamelessly add it to your pull request here :)

Is it possible to share a public link for a task list? I've seen it's possible in the calendar UI, not sure if I'm missing something in the tasks app. The use case is: I'd love to invite some Non-Nextcloud users to a given task list where they at least can see what's on it using their favourite task-app (Apple Reminders, Web UI, or whatever...)

raimund-schluessler:
@marbetschar No, this is not possible yet. I think this issue is tracked in #469. As a workaround you could maybe create a calendar in the Calendar app and publicly share the WebCal link from the Calendar app. Tasks should also show up the in their favourite task-app then (but I haven't tested this).

marbetschar:
@raimund-schluessler understood. Thanks for clarification!

PR #425, opened 2019-05-23
Improve drag and drop experience

raimund-schluessler:
With https://github.com/nextcloud/tasks/pull/418 and the included update to Sortable.js 1.9.0 (specifically https://github.com/SortableJS/Sortable/issues/1487), we don't need to force the subtasks container to be visible in order to drop a task as the only subtask of a parent task.

Hence, there is no empty slot below a hovered task anymore, which makes the drag-and-drop experience less "janky" (@skjnldsv https://github.com/nextcloud/tasks/pull/236#issuecomment-489620659) 😉

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -458,10 +458,6 @@
 		background-color: rgba( $color-primary, .3 );
 	}
 
-	&.dragover > div.subtasks-container > ol {
-		min-height: 37px;
-	}
-
 	.subtasks-container {
 		margin-left: 35px;
 
--- a/src/components/TaskDragContainer.vue
+++ b/src/components/TaskDragContainer.vue
@@ -26,7 +26,7 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 		:move="onMove"
 		@end="onEnd"
 	>
-		<slot :move="onMove" />
+		<slot />
 	</draggable>
 </template>
 
@@ -54,16 +54,6 @@ export default {
 		]),
 
 		/**
-		 * Called when a task is dragged.
-		 *
-		 * @param {Object} $event The event which caused the move
-		 */
-		onMove: function($event) {
-			this.cleanUpDragging()
-			$event.related.classList.add('dragover')
-		},
-
-		/**
 		 * Called when a task is dropped.
 		 *
 		 * @param {Object} $event The event which caused the drop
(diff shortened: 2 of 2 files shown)


skjnldsv:
Awesome!! Works nicely!

Another note though, I feel like moving to a subtask (is it the proper word) is hard to discover. I though it was a ui glitch (shifted like 30px to the right), but it was actually the nesting feature. Any idea how we could do that? :)

raimund-schluessler:
> Another note though, I feel like moving to a subtask (is it the proper word) is hard to discover. I though it was a ui glitch (shifted like 30px to the right), but it was actually the nesting feature. Any idea how we could do that? :)

@skjnldsv No, not really. But to me, this feels natural anyway and I think showing it the same way as it will be shown once it's dropped should be fine. We could maybe highlight the parent-task-to-be as well, but I am not sure as this might lead to different confusion. Also, I am not aware of a different app supporting subtasks like this, so I can't look how others do it. Anyway, let's hear what @jancborchardt thinks 😄

raimund-schluessler:
Also @nextcloud/vue for opinions.

jancborchardt:
Difficult question – it’s always good to look at how different established apps do this.

> Also, I am not aware of a different app supporting subtasks like this, so I can't look how others do it.

If they don’t do it exactly like this, maybe it’s good to look at _how_ they do it, and if there would be a better way or if there’s maybe one established pattern that mgiht be better? :)

raimund-schluessler:
> Difficult question – it’s always good to look at how different established apps do this.
> 
> > Also, I am not aware of a different app supporting subtasks like this, so I can't look how others do it.
> 
> If they don’t do it exactly like this, maybe it’s good to look at _how_ they do it, and if there would be a better way or if there’s maybe one established pattern that mgiht be better? :)

I had a look around and only found one App that does support nested tasks as we do, Todoist. With that I mean, that you can add every property you can add to a task to a subtask, i.e. subtasks are just tasks as well. Many other implementations of subtasks are simply lists of checkpoints in the details field of the task, but not full featured tasks itself.

Regarding drag-and-drop Todoist does it very similar as we do it here, they simply show the subtask-to-be at the new position (see https://get.todoist.help/hc/de/articles/206432369-Teilaufgaben):

Todoist subtasks

I still think, that this is the way to go. In case you don't agree or have an idea how to improve it further, I would propose to open a new issue for this. This PR is still an improvement over the current situation and worth merging 😉 Follow-up PRs are always welcome, though.

jancborchardt:
Sounds good! :) Probably then it’s good for you to keep an eye on how Todoist does things, as their designers will have faced and hopefully solved similar challenges. :)

PR #426, opened 2019-05-23
Comply with our npm guidelines

skjnldsv:
To have the same scripts name between all of our apps and server.
- Running a npm script should only do what you're asking, no subtasks (svg_sprite)
- Unify the naming according to our [convention](https://docs.nextcloud.com/server/16/developer_manual/app/npm.html) 
- Add the NODE_ENV to avoid vuejs issues


skjnldsv:
@raimund-schluessler `make watch-js` ;)

Issue #83, opened 2017-06-01
Subtasks don't display properly in "Today" and "Week" Smart Collections


tschiu:
### Steps to reproduce
1.  Create Task A without a due date
2.  Create Task A1 as a subtask in Task A with due date of Today (or Tomorrow)
3.  Create Task A2 as a subtask of Task A with due date a month from today
4.  Create Task B with due date of Today
5.  Create Task B2 with due date a month from today
6.  Click on Today or Week Smart Collection icon

### Expected behaviour
1.  I would expect Task A not to appear under either Today or Week since there's no due date.
2.  I would expect that Task A1 would appear under both Today and Week since it's got a due date of Today (or Tomorrow)
3.  I would not expect Task A2 to appear under either Today or Week since the due date is neither Today nor Tomorrow
4.  I would expect Task B to appear with under either Today or Week
5.  I would not expect Task B2 to appear under either Today or Week

### Actual behaviour
Task A1 doesn't appear when Task A has no due date set.
If Task A is set with a due date (Today or within a week), then Task A and Task A1 will show up
Task B appears along with Task B2

I suspect this is a database query condition that assumes that the top-level task will have a due date set and doesn't cover the possibility of a subtask having a different due date than the parent task (or that the subtask may have a due date later than the parent task... in which case I would consider it reasonable that the parent task's due date is automatically moved to the latest of the subtasks)

One work-around to get it to display properly is to have the query return the nearest future due date of a subtask if the parent task doesn't have a specified due date, or if the subtask's due date is earlier than the task's due date.

### Server configuration
**Operating system**: Linux tf1-owncloud9 4.4.0-62-generic #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017 x86_64

**Web server:** Apache/2.4.18 (Ubuntu) (apache2handler)

**Database:** mysql 5.1.56

**PHP version:** 7.0.15-0ubuntu0.16.04.4
[...]

PR #431, opened 2019-05-27
Properly show all tasks matching a collection

raimund-schluessler:
This PR aims at solving an issue with the current way we sort tasks into collections (see e.g #83). At the moment we only show a task in a collection if it is a root task and matches the collection in question.

E.g. we only show a task in the `today` collection when it is due today and if it is a root task. In case a subtask is due today, it will not be shown unless its root task is also due today.

The problem here is, that you potentially won't see subtasks due today / this week, are important. This means the collections don't catch all tasks that they should.

With this PR, we show all root tasks that have a subtask matching the collection (or match the collection itself).

The question here is, whether this is reasonable and how we should count the tasks then (for showing their count in the sidebar). Options are to
- only count root tasks (currently done)
- or to count all tasks directly matching the collection)

Feedback from @nextcloud/designers would be very welcome.


jancborchardt:
Do you have screenshots of a scenario? That would be much appreciated to illustrate the issue & solution. :)

raimund-schluessler:
> 
> 
> Do you have screenshots of a scenario? That would be much appreciated to illustrate the issue & solution. :)

This is an example for the `starred` collection which shows important tasks, but it works for the other collections just the same way.

Before this PR the starred collection looks like this:
Screenshot_2019-05-29 Aufgaben - Nextcloud(1)
It shows two important tasks, which are shown because they are important and root tasks. What it does not show is, that there are two more important tasks which are subtasks of an other task.

With this PR, also important subtasks show up in the `starred` collection:
Screenshot_2019-05-29 Aufgaben - Nextcloud
Now the important subtasks lead to the task `Test 5` being shown, also showing the important subtasks. I think, this is the better behaviour, since you now really see all important tasks (subtasks or not).

What I will add is a counter showing the number of sub(sub..)tasks matching this collection (e.g. in a colored circle). So the task `Test 5` will have a counter `2` , the task `Test 5 - Subtask 1` will show `2` as well and `Test 5 - Subtask 1 - Subtask 2` will show  `1`.

raimund-schluessler:
One more question would be now, what the counter in the app-sidebar shows. It could show `3` as it does at the moment with this PR (second screenshot), countin all root tasks. Or it could show the actual number of important tasks, namely `4` in that case.

skjnldsv:
I would say (if I got this right), for the task 5, always show the full task tree if it leads to a starred task.
But don't show any sibling/children tasks that does not contain nor are favourited :)

raimund-schluessler:
> I would say (if I got this right), for the task 5, always show the full task tree if it leads to a starred task.
> But don't show any sibling/children tasks that does not contain nor are favourited :)

That would be an option as well.

jancborchardt:
Hmm, so when you favorited a child task at the 3rd level deep, why not just show only that? That would simplify the whole structure a lot and not make the tree structure necessary. Viewing favorites is about focusing, and the structure might detract.

Otherwise we can do it like you and @skjnldsv proposed – show the structure for reference, but leave out siblings which are not favorited.

raimund-schluessler:
I think not showing the full tree might be confusing. You expect a subtask to be shown as a subtask.

I will implement it as @skjnldsv proposed it. Show the tree if it leads to a starred subtask. I will then show the number of other (not shown) subtasks as a number with a tooltip. When you open the details of this task showing all subtasks then might be good as well.

raimund-schluessler:
Now that it can happen that a task is shown multiple times in the week view (e.g. a task a two subtasks due today and tomorrow) a task should only be shown as active if it is the very exact task.

- [ ] If a task is opened on day 1, don't also show it open on day 2.

raimund-schluessler:
@skjnldsv @jancborchardt This was quite a bit of work, but I think it is finally done.

With this PR all tasks (including subtasks) are shown if they match a collection. We show the full tree for subtasks matching the collection, but don't show siblings which don't match. When a task is open in the details view, all its subtasks are shown.

For the starred view this looks like this.

No task open in details view:
Screenshot_2019-11-14 Aufgaben - Nextcloud(3)


Task which is not matching the collection is open (showing all its subtasks):
Screenshot_2019-11-14 Aufgaben - Nextcloud

Subtask which doesn't match the collection is open (still showing full tree to this subtask):
Screenshot_2019-11-14 Aufgaben - Nextcloud(2)

I also added a lot of tests to check for the correct behaviour of the collections.

What is still left to be done is to indicate which task exactly matches a collection and which one is only shown because a subtask matches. I would like to do this in a follow up PR, because this one is already quite huge.

src/components/TheRouter.js
```diff
@@ -35,7 +35,18 @@ const routes = [
 	// reliably with router-link due to
```
skjnldsv: the router file is usually located into `/src/router/index.js` ;)

raimund-schluessler: Doesn't seem so usual to me 😉 

https://github.com/nextcloud/calendar/blob/master/src/router.js
https://github.com/nextcloud/notes/blob/master/src/router.js
https://github.com/nextcloud/bookmarks/blob/master/src/router.js

vs.

https://github.com/nextcloud/contacts/blob/master/src/router/index.js

But I agree, it is never else located in `src/components/TheRouter.js`. How should we do it?

skjnldsv: I would say it's fine as well in a router.js file, more clear ;)

I like having everything in separate folders, but feel free to choose :)
Both are fine imho (just not in components :wink:  )

vozeldr:
I'm using 0.11.3 on 17.0.1 and this is still an issue, so it doesn't seem like it fixed the problem. Creating a subtask with a due date when the parent does not have a due date does not show that task in Important, Today, or Week.

raimund-schluessler:
This fix is not included in 0.11.3. It will be released with 0.12.0.

PR #450, opened 2019-06-12
Setup Javascript and PHP tests and badges

raimund-schluessler:


raimund-schluessler:
@skjnldsv Do you have an idea why the coverage is not shown on codecov and the badge shows `invalid response data`? I was hoping it would show up after merging into `master` but it does not. The upload works fine and I can see the coverage on https://codecov.io/gh/nextcloud/tasks/commits, but the overview is empty?!

raimund-schluessler:
> @skjnldsv Do you have an idea why the coverage is not shown on codecov and the badge shows `invalid response data`? I was hoping it would show up after merging into `master` but it does not. The upload works fine and I can see the coverage on https://codecov.io/gh/nextcloud/tasks/commits, but the overview is empty?!

Does someone from @nextcloud/javascript has an idea maybe?

skjnldsv:
No idea! :/
I'm not a coverage expert :'(

Issue #506, opened 2019-07-16
Adjust checkbox design to Nextcloud


jancborchardt:
We should adjust the checkboxes in Tasks to look like the core Nextcloud ones. :)

Compare:
![checkbox tasks](https://user-images.githubusercontent.com/925062/61304621-bf499580-a7e9-11e9-8e0b-51c9dbcbdfdf.png) ![Checkbox nextcloud](https://user-images.githubusercontent.com/925062/61304619-bf499580-a7e9-11e9-926e-699125f76c27.png)

See https://twitter.com/despens/status/1151098437875699719 by @despens

Documentation for how to do Nextcloud-style checkboxes are at https://docs.nextcloud.com/server/latest/developer_manual/design/html.html#checkboxes-and-radios

cc @nextcloud/tasks @nextcloud/designers

raimund-schluessler:
I don't know, this looks really weird to me:
![Screenshot-2019-7-16 Aufgaben - Nextcloud](https://user-images.githubusercontent.com/2496460/61306486-ff5e4780-a7ec-11e9-8d7f-ecdabc7d6ae5.png)

The checkbox is to small and it doesn't use the Tasks app icon anymore as the checkmark.

jancborchardt:
The size of the checkbox is adjustable, right @skjnldsv @juliushaertl?

And then we can adjust the Tasks app icon too. :)

raimund-schluessler:
> And then we can adjust the Tasks app icon too, being a checkmark in a box. I can do that. :)

My argument with the same icon does not work in this direction 😉 I don't think we should change the app icon. The icon is
- clear, simple and friendly
- unambiguously
- used for many years

If you really think we have to adjust the checkboxes, I can live with it. I just wonder if the nextcloud design checkboxes might be understood as selecting a task rather than completing it.

despens:
I'd appreciate if the look of checkboxes was unified across Nextcloud apps. 

Users see dozens of differently styled checkboxes all the time, in different desktop and web apps, not even to mention mobile apps. If it looks remotely like a checkbox they will assume it is a checkbox. :wink: 

Also, there are tons of apps that use checkboxes for completing tasks, so for usability the exact styling is not an issue. But it would just be nice if at least within Nextcloud the visual clutter could be reduced and user interface elements provide the same functionality also look the same.

juliushaertl:
[...]

PR #572, opened 2019-08-29
first try to implement the nextcloud checkbox design

jakobroehrl:
Signed-off-by: Jakob <jakob.roehrl@web.de>

@raimund-schluessler: Cloud you help me here with the CSS?


raimund-schluessler:
@jakobroehrl I haven't had time to check the functionality or the CSS yet, but I still have some feedback already:

- [ ] The ID of an element has to be unique on the whole page. You have to add a unique identifier to the ID if you use it in a reusable component. Just append `task.uid` to your ID. `:id="' toggleCompleted_' + task.uid" should work like here https://github.com/nextcloud/tasks/blob/master/src/components/TheList.vue#L52
- [ ] The checkbox has to be disabled in case the calendar is read-only: `task.calendar.readOnly`.
- [ ] The checkbox style is used for the toggle all day checkbox in the details view as well: https://github.com/nextcloud/tasks/blob/master/src/components/TheDetails.vue#L132-L137. It should be adjusted too.
- [ ] The aria attributes should be kept. Just add them to the input element.

Bonus:

- [ ] The checkmark icon is used as indicator to show a successful sync to the server. Once we have the final design, we should adjust the icon as well: https://github.com/nextcloud/tasks/blob/master/img/src/color/checkmark-color.svg

jakobroehrl:
@raimund-schluessler 
Added your Feedback and the CSS from Julius: https://github.com/nextcloud/tasks/issues/506#issuecomment-526306976

raimund-schluessler:
Should be all good now.

raimund-schluessler:
Before:
Screenshot_2019-09-01 Aufgaben - Nextcloud(1)

After:
Screenshot_2019-09-01 Aufgaben - Nextcloud

PR #591, opened 2019-09-06
Use replace built into translate function

raimund-schluessler:
Signed-off-by: Raimund Schlüßler <raimund.schluessler@mailbox.org>


raimund-schluessler:
Merging this, so the translations can be updated quickly.

Issue #566, opened 2019-08-27
Tasks App is blank in Chrome Beta 77


mcocolla:
### Steps to reproduce
1. Install Google Chrome Beta 77
2. Open Tasks 
3. No tasks show

### Expected behaviour
Tasks lists and tasks from those lists should show on the screen.

### Actual behaviour
JavaScript throws 3 errors from tasks.js and the screen is blank. Add list does not work either.
Errors are listed below in the Browser log section. 

Stable versions of Chrome, Firefox, IE11, and Edge currently work as expected. Not sure if it is a bug in Chrome Beta or if it is a change to the JavaScript engine that is permanent. 

### Client configuration
**Browser:** Chrome Beta 77.0.3865.42

**Operating system:** Windows 10 1903

#### Browser log
```
Insert your browser log here, this could for example include:

tasks.js?v=5d618559-11:7 TypeError: t(...).replace is not a function
    at a.deleteMessage (tasks.js?v=5d618559-11:845)
    at tasks.js?v=5d618559-11:845
    at a.ge [as _l] (tasks.js?v=5d618559-11:7)
    at a.<anonymous> (tasks.js?v=5d618559-11:845)
    at a.t._render (tasks.js?v=5d618559-11:7)
    at a.r (tasks.js?v=5d618559-11:7)
    at hn.get (tasks.js?v=5d618559-11:7)
    at hn.run (tasks.js?v=5d618559-11:7)
    at dn (tasks.js?v=5d618559-11:7)
    at Array.<anonymous> (tasks.js?v=5d618559-11:7)
tasks.js?v=5d618559-11:7 TypeError: t(...).replace is not a function
    at a.inputString (tasks.js?v=5d618559-11:845)
    at hn.get (tasks.js?v=5d618559-11:7)
    at hn.evaluate (tasks.js?v=5d618559-11:7)
    at a.inputString (tasks.js?v=5d618559-11:7)
    at a.<anonymous> (tasks.js?v=5d618559-11:845)
    at a.t._render (tasks.js?v=5d618559-11:7)
    at a.r (tasks.js?v=5d618559-11:7)
    at hn.get (tasks.js?v=5d618559-11:7)
    at hn.run (tasks.js?v=5d618559-11:7)
    at dn (tasks.js?v=5d618559-11:7)
tasks.js?v=5d618559-11:7 TypeError: t(...).replace is not a function
    at a.deleteMessage (tasks.js?v=5d618559-11:845)
[...]

PR #593, opened 2019-09-06
Fix translate function for Chrome 77 Beta

raimund-schluessler:
This fixes an issue occuring in Chrome 77 Beta, see #566. Although this would fix the issue and not cause any other problems (as far as I know atm), I am a bit reluctant to actually merge this because:

- we fix an issue occuring in the Beta release of Chrome which might not occur anymore in the stable release
- in case the issue persists in the Chrome stable release, the translate function `t(...)` should actually be adjusted in Nextcloud server and not in the apps.

As a conclusion, I will for now not merge this fix but wait how things go with Chrome stable. In case it persists, we have a fix at hand and can react quickly. For everyone that wants/has to use Chrome Beta here is a version of the app containing the fix:
[tasks.zip](https://github.com/nextcloud/tasks/files/3595292/tasks.zip)

### DIFF
--- a/src/components/Task.vue
+++ b/src/components/Task.vue
@@ -157,9 +157,9 @@ export default {
 		formatDate: function(date) {
 			return valid(date)
 				? moment(date, 'YYYYMMDDTHHmmss').calendar(null, {
-					lastDay: t('tasks', '[Yesterday]'),
-					sameDay: t('tasks', '[Today]'),
-					nextDay: t('tasks', '[Tomorrow]'),
+					lastDay: OCA.Tasks.$t('tasks', '[Yesterday]'),
+					sameDay: OCA.Tasks.$t('tasks', '[Today]'),
+					nextDay: OCA.Tasks.$t('tasks', '[Tomorrow]'),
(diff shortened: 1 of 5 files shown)


mcocolla:
Chrome 77 just went stable release and in my testing the problem persists with the translate function. I've also noticed placeholder text on modal dialogs in 77, not sure if that is also part of the same root problem. 

Annotation 2019-09-10 144127

raimund-schluessler:
> Chrome 77 just went stable release and in my testing the problem persists with the translate function. I've also noticed placeholder text on modal dialogs in 77, not sure if that is also part of the same root problem.

Indeed, I can confirm that Chrome 77 produces this error. Since there is no solution in server yet, let's merge this PR to fix the problem in the Tasks app. We can then still react to whatever changes will be introduced to server.

raimund-schluessler:
@mcocolla Can you elaborate on this modal dialog error? How did you produce this modal?

mcocolla:
@raimund-schluessler This dialog came from the settings > basic settings and clicking the store credentials button on the Email server settings.

Issue #606, opened 2019-09-11
Try to load task from server if not found locally


raimund-schluessler:
When a task is not found on the client side, it should be queried from the server. At the moment a task is shown as `Not found` when it is opened, but not loaded from the server (this can happen e.g. for completed tasks, since they are not loaded by default from the server).

georgehrke:
we need some special `getEntityAtUrl` method. Let me look into that today.

georgehrke:
If you already have the cdav-calendar object, you can just use the find method btw.
https://github.com/nextcloud/cdav-library/blob/master/src/models/davCollection.js#L111

raimund-schluessler:
> If you already have the cdav-calendar object, you can just use the find method btw.
> https://github.com/nextcloud/cdav-library/blob/master/src/models/davCollection.js#L111

Thanks, I must have overlooked this method. It should work fine.

raimund-schluessler:
PR is in #608.

raimund-schluessler:
Closed with #608.

PR #608, opened 2019-09-13
Load task from server if not found locally

raimund-schluessler:
This implements #606. When a requested task is not found locally, we try to load it from the server.

Todo:
- [x] We should also try to load the ancestor tasks (if any) of the requested task, so we can properly show the subtasks tree
- [x] If we are in a collection view (e.g. /nextcloud/index.php/apps/tasks/#/collections/all/tasks/Nextcloud-lrr0rvhdcoqopte5k43n6.ics), we don't know which calendar a task belongs to, so we have to query all calendars I guess. Or is there a better way to find a task only knowing the last part of its `uri` @georgehrke @skjnldsv ?
- [x] Don't loose subtask relations for tasks alread present in store when loading tasks by Uri or UID

What I don't like so much is that I have to watch the calendar (https://github.com/nextcloud/tasks/pull/608/files#diff-128fe9ee0ed5aeabe3c64281c67285c0R584), but it was the easiest way to load a task only after the calendars are loaded.

### DIFF
--- a/src/components/TheDetails.vue
+++ b/src/components/TheDetails.vue
@@ -21,7 +21,7 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 
 <template>
 	<div class="content-wrapper">
-		<div v-if="task!=undefined"
+		<div v-if="task"
 			:class="{'disabled': task.calendar.readOnly}"
 			class="flex-container"
 		>
@@ -340,7 +340,8 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 			</div>
 		</div>
 		<div v-else class="notice">
-			<span>{{ $t('tasks', 'Task not found!') }}</span>
+			<span v-if="loading">{{ $t('tasks', 'Loading task from server.') }}</span>
+			<span v-else>{{ $t('tasks', 'Task not found!') }}</span>
 		</div>
 	</div>
 </template>
@@ -500,6 +501,7 @@ export default {
 	},
 	data: function() {
 		return {
+			loading: false,
 			edit: '',
 			tmpTask: {
 				summary: '',
@@ -572,8 +574,19 @@ export default {
 		},
 		...mapGetters({
 			writableCalendars: 'getSortedWritableCalendars',
-			task: 'getTaskByRoute'
+			task: 'getTaskByRoute',
+			calendar: 'getCalendarByRoute',
+			calendars: 'getSortedCalendars',
 		}),
+	},
+
+	watch: {
+		$route: 'loadTask',
+		calendars: 'loadTask',
+	},
+
+	created() {
+		this.loadTask()
 	},
 
 	/**
@@ -615,7 +628,27 @@ export default {
 			'toggleAllDay',
 			'moveTask',
 			'setClassification',
+			'getTaskByUri',
 		]),
+
+		async loadTask() {
+			if (this.task === undefined || this.task === null) {
+				const calendars = this.calendar ? [this.calendar] : this.calendars
+				for (const calendar of calendars) {
+					this.loading = true
+					try {
+						const task = await this.getTaskByUri({ calendar, taskUri: this.$route.params.taskId })
+						// If we found the task, we don't need to query the other calendars.
+						if (task) {
+							break
+						}
+					} catch {
+						console.debug('Task ' + this.$route.params.taskId + ' not found in calendar ' + calendar.displayName + '.')
+					}
+				}
+				this.loading = false
+			}
+		},
 
 		removeTask: function() {
 			this.deleteTask({ task: this.task, dav: true })
--- a/src/store/tasks.js
+++ b/src/store/tasks.js
@@ -27,6 +27,7 @@ import { isParentInList, momentToICALTime } from './storeHelper'
 import ICAL from 'ical.js'
 import TaskStatus from '../models/taskStatus'
 import router from '../components/TheRouter'
+import { findVTODObyUid } from './cdav-requests'
 
 Vue.use(Vuex)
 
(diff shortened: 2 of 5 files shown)


src/components/TheDetails.vue
```diff
@@ -630,17 +631,30 @@ export default {
...
-						this.loading = false
-						throw error
-					})
+		async loadTask() {
+			if (this.task === undefined || this.task === null) {
+				const calendars = this.calendar ? [this.calendar] : this.calendars
+				for (const calendar of calendars) {
+					this.loading = true
+					const found = await this.getTaskByUri({ calendar: calendar, taskUri: this.$route.params.taskId })
+						.then((response) => {
```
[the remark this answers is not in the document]
raimund-schluessler: So, like this 5615faf58c8ddfc7a3cdb12c045f6dcba23eca8c?

raimund-schluessler:
This should be good to go now.

src/components/TheDetails.vue
```diff
@@ -615,8 +628,36 @@ export default {
...
 		]),
 
+		async loadTask() {
+			if (this.task === undefined || this.task === null) {
+				const calendars = this.calendar ? [this.calendar] : this.calendars
+				for (const calendar of calendars) {
+					this.loading = true
+					try {
+						const found = await this.getTaskByUri({ calendar: calendar, taskUri: this.$route.params.taskId })
+							.then((response) => {
```
skjnldsv: please don't mix try..;catch and .then/catch :)

src/store/tasks.js
```diff
@@ -720,6 +721,96 @@ const actions = {
...
+	 * and commits the result
+	 *
+	 * @param {Object} context The store mutations
+	 * @param {Object} data Destructuring object
+	 * @param {Calendar} data.calendar The calendar
+	 * @param {String} data.taskUri The uri of the requested task
+	 * @returns {Promise}
+	 */
+	async getTaskByUri(context, { calendar, taskUri }) {
+		return calendar.dav.find(taskUri)
```
skjnldsv: try...catch? :)

src/components/TheDetails.vue
raimund-schluessler: What would be the correct way? Could you elaborate a bit, I am not really familiar with this.

skjnldsv: It's usually good practice to not use .then/.catch anymore as you end up with chaining lots of promises :)
Using async/await is recommended ;)

raimund-schluessler: Ok, I will have a look. Thanks!

raimund-schluessler: Thanks for making me aware @skjnldsv. Looks much cleaner with async/await only 👍

raimund-schluessler:
> See comments :)

Fixed it. Looks much better now with async/await.

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
 
+	/**
+	 * Retrieves the task with the given uri from the given calendar
+	 * and commits the result
+	 *
+	 * @param {Object} context The store mutations
+	 * @param {Object} data Destructuring object
+	 * @param {Calendar} data.calendar The calendar
+	 * @param {String} data.taskUri The uri of the requested task
+	 * @returns {Promise}
```
skjnldsv:
```suggestion
	 * @returns {Task}
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+
+	/**
+	 * Retrieves the task with the given uid from the given calendar
+	 * and commits the result
+	 *
+	 * @param {Object} context The store mutations
+	 * @param {Object} data Destructuring object
+	 * @param {Calendar} data.calendar The calendar
+	 * @param {String} data.taskUid The uid of the requested task
+	 * @returns {Promise}
```
skjnldsv:
```suggestion
	 * @returns {Task}
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+		if (response) {
+			const task = new Task(response.data, calendar)
+			Vue.set(task, 'dav', response)
+			if (task.related) {
+				let parent = context.getters.getTaskByUid(task.related)
+				// If the parent is not found locally, we try to get it from the server.
+				if (!parent) {
+					parent = await context.dispatch('getTaskByUid', { calendar: calendar, taskUid: task.related })
+				}
+				context.commit('addTaskToParent', { task: task, parent: parent })
```
skjnldsv:
```suggestion
				context.commit('addTaskToParent', { task, parent })
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+	async getTaskByUri(context, { calendar, taskUri }) {
+		const response = await calendar.dav.find(taskUri)
+		if (response) {
+			const task = new Task(response.data, calendar)
+			Vue.set(task, 'dav', response)
+			if (task.related) {
+				let parent = context.getters.getTaskByUid(task.related)
+				// If the parent is not found locally, we try to get it from the server.
+				if (!parent) {
+					parent = await context.dispatch('getTaskByUid', { calendar: calendar, taskUid: task.related })
```
skjnldsv:
```suggestion
					parent = await context.dispatch('getTaskByUid', { calendar, taskUid: task.related })
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+
+			// In case we already have subtasks of this task in the store, add them as well.
+			const subTasksInStore = context.getters.getTasksByParent(task)
+			subTasksInStore.forEach(
+				subTask => {
+					context.commit('addTaskToParent', { task: subTask, parent: task })
+				}
+			)
+
+			context.commit('appendTasksToCalendar', { calendar: calendar, tasks: [task] })
```
skjnldsv:
```suggestion
			context.commit('appendTasksToCalendar', { calendar, tasks: [task] })
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+		const response = await findVTODObyUid(calendar, taskUid)
+		// We expect to only get zero or one task when we query by UID.
+		if (response.length) {
+			const task = new Task(response[0].data, calendar)
+			Vue.set(task, 'dav', response[0])
+			if (task.related) {
+				let parent = context.getters.getTaskByUid(task.related)
+				// If the parent is not found locally, we try to get it from the server.
+				if (!parent) {
+					parent = await context.dispatch('getTaskByUid', { calendar: calendar, taskUid: task.related })
```
skjnldsv:
```suggestion
					parent = await context.dispatch('getTaskByUid', { calendar, taskUid: task.related })
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+				}
+				context.commit('addTaskToParent', { task: task, parent: parent })
```
skjnldsv:
```suggestion
				context.commit('addTaskToParent', { task, parent })
```

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+
+			// In case we already have subtasks of this task in the store, add them as well.
+			const subTasksInStore = context.getters.getTasksByParent(task)
+			subTasksInStore.forEach(
+				subTask => {
+					context.commit('addTaskToParent', { task: subTask, parent: task })
+				}
+			)
+
+			context.commit('appendTasksToCalendar', { calendar: calendar, tasks: [task] })
```
skjnldsv:
```suggestion
			context.commit('appendTasksToCalendar', { calendar, tasks: [task] })
```

src/components/TheDetails.vue
```diff
@@ -615,8 +628,28 @@ export default {
...
+			'getTaskByUri',
 		]),
 
+		async loadTask() {
+			if (this.task === undefined || this.task === null) {
+				const calendars = this.calendar ? [this.calendar] : this.calendars
+				for (const calendar of calendars) {
+					this.loading = true
+					try {
+						const task = await this.getTaskByUri({ calendar: calendar, taskUri: this.$route.params.taskId })
```
skjnldsv: Much cleaner ! :+1: 
```suggestion
						const task = await this.getTaskByUri({ calendar, taskUri: this.$route.params.taskId })
```

src/store/tasks.js
raimund-schluessler: Although strictly async functions return Promises as well 😉

raimund-schluessler:
Rebased.

src/store/tasks.js
```diff
@@ -720,6 +721,88 @@ const actions = {
...
+	 * @param {Object} data Destructuring object
+	 * @param {Calendar} data.calendar The calendar
+	 * @param {String} data.taskUri The uri of the requested task
+	 * @returns {Task}
+	 */
+	async getTaskByUri(context, { calendar, taskUri }) {
+		const response = await calendar.dav.find(taskUri)
+		if (response) {
+			const task = new Task(response.data, calendar)
+			Vue.set(task, 'dav', response)
```
skjnldsv: Not a big deal, more a small tip, but you can also use `this.$set` :)

skjnldsv: Ah no, we're in a store! Sorry!! :grin: :wink:

PR #656, opened 2019-10-08
save all day Setting

jakobroehrl:
Signed-off-by: Jakob <jakob.roehrl@web.de>

This is the first try. 
I have 2 problems: 
1. setting is saved but lost on page reload
2. setting is not applied on new created tasks

@raimund-schluessler: Cloud you give me some hints? Thanks!

### DIFF
--- a/src/components/TheCollections/Calendar.vue
+++ b/src/components/TheCollections/Calendar.vue
@@ -152,7 +152,21 @@ export default {
 		},
 
 		addTask: function() {
-			this.createTask({ summary: this.newTaskName, calendar: this.calendar })
+			if (this.$store.state.settings.settings.defaultAllDay) {
+
+				this.createTask({
+					summary: this.newTaskName,
+					calendar: this.calendar,
+					allDay: true
+				})
+			} else {
+				this.createTask({
+					summary: this.newTaskName,
+					calendar: this.calendar
+				})
+
+			}
+
 			this.newTaskName = ''
 		},
 	}
--- a/src/store/tasks.js
+++ b/src/store/tasks.js
@@ -592,6 +592,9 @@ const actions = {
 		if (taskData.start) {
 			task.start = taskData.start
 		}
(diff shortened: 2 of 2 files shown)


raimund-schluessler:
This is a tricky one. I also started working on it locally and implemented saving the setting to the server. But I didn't push it yet, because applying the setting for a new date requires more work. Anyway, it might be good to use my branch, because saving the setting is already there. I will push it asap.

The real problem is applying the setting when you first add a date to a task. The allDay toggle is not a value on its own, but is determined by the date that is currently set for the task: https://github.com/nextcloud/tasks/blob/master/src/models/task.js#L365-L381
When you change it, the date is toggled between allday and with-time, and then the value for allDay is determined from it. The problem now is, that you cannot set a value for allday while no date is set. So changing the value of allDay without a date present will lead to nothing.

I don't have a good solution for this yet, but it will definitely require a bit of refactoring.

jakobroehrl:
> 
> 
> This is a tricky one. I also started working on it locally and implemented saving the setting to the server. But I didn't push it yet, because applying the setting for a new date requires more work. Anyway, it might be good to use my branch, because saving the setting is already there. I will push it asap.
> 
> The real problem is applying the setting when you first add a date to a task. The allDay toggle is not a value on its own, but is determined by the date that is currently set for the task: https://github.com/nextcloud/tasks/blob/master/src/models/task.js#L365-L381
> When you change it, the date is toggled between allday and with-time, and then the value for allDay is determined from it. The problem now is, that you cannot set a value for allday while no date is set. So changing the value of allDay without a date present will lead to nothing.
> 
> I don't have a good solution for this yet, but it will definitely require a bit of refactoring.

Thank you @raimund-schluessler So, I will wait, till you merged you changes...

raimund-schluessler:
Superseded by #657. @jakobroehrl Feel free to push to #657.

Issue #637, opened 2019-09-24
Set a personal start view


jakobroehrl:
It would be nice to set the start view: Important, Today, Week, All
Currently tasks starts always with the All view.

Btw: What's the difference between All and Current?
![grafik](https://user-images.githubusercontent.com/24530680/65504500-6ff9a280-dec7-11e9-8df3-fd73ec216591.png)

raimund-schluessler:
Duplicate of #80. And I am not very keen on introducing a new setting.
But we could remember the last view, similar as in #615.

And the Current view does not show tasks which have not begun yet (i.e. a start date in the future).

Issue #80, opened 2017-05-24
Remember last view


borghal:
I'd like to be able to define the default view when tasks is opened. 

Right now when you open tasks it shows a list of all tasks which is not very practical in my opinion. I'd love to open tasks and see the "today" list first, for example.

Cheers!

jakobroehrl:
Hello @raimund-schluessler 
I would like to do this for the next version. Could you give me the steps for realizing this?

raimund-schluessler:
@jakobroehrl @jancborchardt I don't think we will introduce a setting for this, will we? We could save the last selected view though, similar how we do it for the all-day setting. Although this wouldn't *really* be a default view then.

jancborchardt:
>  We could save the last selected view though

Yes, always remembering the last used view is the proper way to do this. :)

PR #657, opened 2019-10-08
Apply last all-day setting to newly created dates

### LINKED ISSUE #615: [Feature Request] Remember setting of all-day toggle for creation of new tasks
I would love to have the all day option set by default.

raimund-schluessler:
Closes #615

As discussed in https://github.com/nextcloud/tasks/pull/656#issuecomment-539482473 applying the all-day setting to newly created dates is still to be done here.

### DIFF
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -63,6 +63,7 @@ class SettingsService {
 			'showHidden' => (int)$this->settings->getUserValue($this->userId, $this->appName,'various_showHidden'),
 			'sortOrder' => (string)$this->settings->getUserValue($this->userId, $this->appName,'various_sortOrder'),
 			'sortDirection' => (bool)$this->settings->getUserValue($this->userId, $this->appName,'various_sortDirection'),
+			'allDay' => (bool)$this->settings->getUserValue($this->userId, $this->appName,'various_allDay'),
 			'userID' => $this->userId
 		);
 		return $settings;
--- a/src/components/TheCollections/General.vue
+++ b/src/components/TheCollections/General.vue
@@ -154,6 +154,7 @@ export default {
 			}
 			if (this.$route.params.collectionId === 'today') {
 				task.due = moment().startOf('day').format('YYYY-MM-DDTHH:mm:ss')
+				task.allDay = this.$store.state.settings.settings.allDay
 			}
 			if (this.$route.params.collectionId === 'current') {
 				task.start = moment().format('YYYY-MM-DDTHH:mm:ss')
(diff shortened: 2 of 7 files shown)


jakobroehrl:
@raimund-schluessler 
Checked your changes. Saving the settings works. 
Is it possible to show allDay always and remove the connection to the dates?

jakobroehrl:
@raimund-schluessler
Could we solve #637 easy if we copy and modify these change a litte bit?

raimund-schluessler:
> 
> 
> @raimund-schluessler
> Could we solve #80 easy if we copy and modify these change a litte bit?

Well, we could store the last view in the same way.

raimund-schluessler:
@jakobroehrl This should work as expected now. Can you please test?

jakobroehrl:
@raimund-schluessler Thanks for your work. 
Unfortunately I don't see any changes. Added task 1, swiched to all day, added task 2, all day isn't set. 
Did I make a mistake?
Pulled fix-615, npm install, cleared the browser cache.

raimund-schluessler:
> Pulled fix-615, npm install, cleared the browser cache.

`npm install` is not necessary, but you need to build the app with `make` or `make build-js`. If it is built correctly you should see requests for saving the setting to the server when changing the all-day setting.

jakobroehrl:
Yes, I see the network request when I check/uncheck the all-day checkbox.

raimund-schluessler:
> Yes, I see the network request when I check/uncheck the all-day checkbox.

So, does it work now? Did you build the app?

jakobroehrl:
> So, does it work now? Did you build the app?
I build the app.
I see two request after I ticked all-day
But when I create a new task I have to tick all-day every time.

raimund-schluessler:
Hm, I think I didn't check if it works when creating a new task (although I think it should). Could you check whether it remembers the setting when you remove start and due date and set it again. I will check in the evening if it works for creating new tasks.

jakobroehrl:
> Hm, I think I didn't check if it works when creating a new task (although I think it should). Could you check whether it remembers the setting when you remove start and due date and set it again. I will check in the evening if it works for creating new tasks.

That's working. It would be great if it would work for new tasks, too. Thank you!

raimund-schluessler:
> That's working. It would be great if it would work for new tasks, too. Thank you!

Weird. I will check.

raimund-schluessler:
@jakobroehrl It works as it should for me. This is what I do:
- create a new task
- set a start or due date, make it all-day
- create a second task
- set a start or due date, which is already all-day

Could you please describe in detail what you do, so it doesn't work?

The only problem I noticed is that creating tasks in the "Today" collection does not respect the all-day setting yet. I will adjust this.

raimund-schluessler:
Just to make it clear, when you create a task, all-day is only set when you set a start or due date. Without a date, all-day cannot be applied (since it is not a value itself, but determined from the present date).

jakobroehrl:
Now it's working. :)
Restarted the VM und cleared the cache again.

PR #684, opened 2019-10-22
Adjust complete value when setting status

### LINKED ISSUE #663: Resolve status - completion dependency
One thing that just came to my mind: When we change the complete value of a task, whe change the status accordingly: https://github.com/nextcloud/tasks/blob/master/src/models/task.js#L242-L257
complete 0 % => NEEDS-ACTION,
complete > 0 % => IN-PROCESS,
complete 100 % => COMPLETED

I think we should we vice-versa also set the complete value when we adjust the status manually. At the moment it is a bit weird e.g. if you have a completed task and manually set the status to IN-PROCESS. We might want to set the complete value to e.g. 99 % then.

So:
NEEDS-ACTION: complete = 0
IN-PROCESS: complete > 0, < 100,
COMPLETED: complete 100,
CANCELED: leave the complete value as is.

raimund-schluessler:
Closes #663

### DIFF
--- a/src/models/task.js
+++ b/src/models/task.js
@@ -240,19 +240,26 @@ export default class Task {
 	}
 
 	set complete(complete) {
-		this.vtodo.updatePropertyWithValue('percent-complete', complete)
-		this.updateLastModified()
+		// Make complete a number
+		complete = +complete
+		this.setComplete(complete)
 		if (complete < 100) {
-			this.completed = false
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
Tests fail, because #663 is not implemented yet.

raimund-schluessler:
Should be good to go now.

Issue #694, opened 2019-10-30
Add Support for pinning tasks


timholl:
Open Tasks (and maybe some others) use the `X-PINNED` extension to pin a task onto the users homescreen.
I would like to be able to set that task property via the web interface on my computer.

Pull request follows.

raimund-schluessler:
Closed with #695

PR #695, opened 2019-10-30
Added support for task pinning

timholl:
Implements #694


timholl:
Pinned state can be toggled by clicking on the pin icon next to the star

detail

detail2

Pinned tasks also get an additional icon in the list view
list

raimund-schluessler:
Thanks for the contribution.

Regarding the feature:
This feature has no effect in the web interface. Toggling the pinned state does not change anything and only has an effect for users of the OpenTasks app. Considering this, the implementation is quite prominent in the web interface. I wonder whether we can realise this a bit more unobtrusively. @nextcloud/designers @jancborchardt Any opinion?

Regarding the code:
It looks very good and clean 👍 You only need to signoff your commit, see https://github.com/nextcloud/tasks/pull/695/checks?check_run_id=282263610

css/src/style.scss
```diff
@@ -698,7 +698,7 @@
 				display: inline-block;
 				min-height: 24px;
 				margin: 13px 0;
-				width: calc(100% - 116px);
+				width: calc(100% - (41px + 16px + 45px + 45px + 14px)); /* checkbox (41px), status display (16px),  star (45px), pin (45px), safety distance (14px) */
```
raimund-schluessler: This looks bad. I guess we should use flexbox for the title of the details view. Can be done in a follow-up PR, though.

src/components/TheDetails.vue
```diff
@@ -587,6 +589,13 @@ export default {
 				return 'icon-bw icon-task-star'
 			}
 		},
+		iconPinned: function() {
+			if (+this.task.pinned) {
```
skjnldsv:
```suggestion
			if (this.task.pinned) {
```

raimund-schluessler:
We could also try to add some benefit for the web interface. E.g. show a pinned task always at the top of a list.

timholl:
To justify myself a litte for https://github.com/nextcloud/tasks/pull/695/files#diff-88cbe2e916273c5c244516d0ec6b33ffR756-R757 :

I had a little struggle with the icons since I added two and as a side-effect the big star icon (and the new pin) did not work anymore.
I figured out that was because the sprite was resized from 5x5 to 6x5.
No problem for `svg-sprite` module to find the correct positions in CSS, but the 1.5-upscaling of the background for `.star.icon-bw` class was hardcoded in the CSS for 5x5 (to `120px 120px`) and also for `.star.icon-color`. Had to  take that numbers apart and put them humpty-dumpty back together for 6x5 (kinda "improved" it by using `auto` for the second parameter so there is just one hardcoded number left).

Maybe one could find a more conceptual solution for this on the future than incrementing a hardcoded number ;)

src/components/TheDetails.vue
```diff
@@ -587,6 +589,13 @@ export default {
 				return 'icon-bw icon-task-star'
 			}
 		},
+		iconPinned: function() {
+			if (this.task.pinned) {
+				return 'icon-bw icon-pinned-off'
```
raimund-schluessler:
```suggestion
				return 'icon-bw icon-pinned'
```

src/components/TheDetails.vue
```diff
@@ -587,6 +589,13 @@ export default {
...
+			} else {
+				return 'icon-bw icon-pinned'
```
raimund-schluessler:
```suggestion
				return 'icon-bw icon-pinned-off'
```

src/components/TheDetails.vue
raimund-schluessler: I think it is a bit confusing that the opposite icon is shown in the Task body and the details header. And we normally use the icon indicating the current state, and not the state you will achieve clicking the icon (e.g. the starred icon is red for important tasks).

src/components/TheDetails.vue
```diff
@@ -40,7 +40,9 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 					>
 					<label :for="'detailsToggleCompleted_' + task.uid" />
 				</span>
-
+				<a class="star reactive" @click="togglePinned(task)">
+					<span :class="[{'disabled': task.calendar.readOnly}, iconPinned]" class="icon" />
+				</a>
```
raimund-schluessler: Please insert the icon after the star icon. I know it's confusing but since both the star icon and the pinned icon are `float: right` the pinned icon is at the very right now and we should keep the star icon at the very right.

raimund-schluessler:
See my two comments, I think we are good after that.

raimund-schluessler:
> To justify myself a litte for https://github.com/nextcloud/tasks/pull/695/files#diff-88cbe2e916273c5c244516d0ec6b33ffR756-R757 :
> 
> I had a little struggle with the icons since I added two and as a side-effect the big star icon (and the new pin) did not work anymore.
> I figured out that was because the sprite was resized from 5x5 to 6x5.
> No problem for `svg-sprite` module to find the correct positions in CSS, but the 1.5-upscaling of the background for `.star.icon-bw` class was hardcoded in the CSS for 5x5 (to `120px 120px`) and also for `.star.icon-color`. Had to take that numbers apart and put them humpty-dumpty back together for 6x5 (kinda "improved" it by using `auto` for the second parameter so there is just one hardcoded number left).
> 
> Maybe one could find a more conceptual solution for this in the future than incrementing a hardcoded number ^^

Don't worry, I will have a look at this in a follow-up PR. We will move the header to `flex-box` then this should not be necessary anymore.

src/components/TheDetails.vue
timholl: Sorry for that, i copied the behaviour of OpenTasks at that point :)

raimund-schluessler:
> I had a little struggle with the icons since I added two and as a side-effect the big star icon (and the new pin) did not work anymore.
> I figured out that was because the sprite was resized from 5x5 to 6x5.
> No problem for `svg-sprite` module to find the correct positions in CSS, but the 1.5-upscaling of the background for `.star.icon-bw` class was hardcoded in the CSS for 5x5 (to `120px 120px`) and also for `.star.icon-color`. Had to take that numbers apart and put them humpty-dumpty back together for 6x5 (kinda "improved" it by using `auto` for the second parameter so there is just one hardcoded number left).

I only now got what you meant. This is indeed not really nice, but I couldn't find any other solution so far. And obviously, flex-box won't do much about it.

raimund-schluessler:
I tend to reduce the pin size down to a normal icon (16px x 16px), but let's discuss this in a later PR.

PR #697, opened 2019-10-31
Align pin svgs

raimund-schluessler:
I slightly adjusted the pin svgs to align, so the pin does not move when toggling the pinned state. Also, I made them viewable by the browser (e.g. Firefox). @timholl


timholl:
:+1: also tested, looks good.

PR #698, opened 2019-10-31
Don't require knowledge of sprite row count

raimund-schluessler:
@timholl This fixes the problem we talked about in https://github.com/nextcloud/tasks/pull/695#issuecomment-548146561.

The alternative would be to force the sprite to have only one row (should be possible with layout: vertical, see https://github.com/jkphl/svg-sprite/blob/master/docs/configuration.md#specific-mode-properties). But I think `transform: scale(1.5); ` should be fine.

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -774,19 +774,14 @@
 				padding: 10px;
 				height: 44px;
 				width: 44px;
+				display: flex;
+				align-items: center;
+				justify-content: center;
 
 				.icon {
-					height: 24px;
-					width: 24px;
-					&.icon-color {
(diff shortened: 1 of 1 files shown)


timholl:
:+1: Works like a charm.

timholl:
> The alternative would be to force the sprite to have only one row (should be possible with layout: vertical, see https://github.com/jkphl/svg-sprite/blob/master/docs/configuration.md#specific-mode-properties).

I also thought about changing the sprite layout, but as far as I've read, an extra yml would have to be added and would not have been a clean solution anyway :)

Issue #392, opened 2019-05-08
Respect minimum clickable area of 44px


skjnldsv:
Every item that can be clicked or selected needs to have a minimum size of 44x44. :wink:
You can use a scss variable like we do on the Vue components. It allows an easy use of the 44px size. (See `&clickable-area`)

raimund-schluessler:
Closed with #735

PR #735, opened 2019-11-26
Increase clickable area of active elements

raimund-schluessler:
This PR adjusts the clickable area of active elements to be 44px x 44px following the Nextcloud standard and closing #392. @skjnldsv @jancborchardt 
The sortorder dropdown in the top right I will adjust in a follow-up PR.

Before:
Screenshot_2019-11-26 Aufgaben - Nextcloud(1)

After:
Screenshot_2019-11-26 Aufgaben - Nextcloud

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -996,23 +996,7 @@
 					}
 
 					&.section-utils button {
-						box-sizing: border-box;
-						width: 44px;
-						height: 44px;
-						margin: 0;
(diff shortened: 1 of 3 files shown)


jancborchardt:
Very cool @raimund-schluessler! :) I’d say another thing which would really improve the look of the app is thinking about ways to remove all the border lines (for that we mainly have to think about how to present the info presented as icons on the right).

PR #747, opened 2019-12-04
Allow to add tasks due today in week view

### LINKED ISSUE #742: Add create tasks filed to week view
It would be nice if tasks can create in the week view, too.
I suggest to create them in the same way as in the today view, with due of today.

If it's ok, I could try to implement it.

raimund-schluessler:
This PR allows to add tasks which are due in a week (today + 7) in the week view. I adjusted the week view to show today + 7 days (8 days in total), instead of 7 days so the newly created tasks actually show up in week view.

Closes #742.

@jakobroehrl

### DIFF
--- a/src/components/TheCollections/Week.vue
+++ b/src/components/TheCollections/Week.vue
@@ -114,7 +114,7 @@ export default {
 		},
 
 		inputString: function() {
-			return this.$t('tasks', 'Add a task due in one week to "{calendar}"…', { calendar: this.calendar.displayName })
+			return this.$t('tasks', 'Add a task due today to "{calendar}"…', { calendar: this.calendar.displayName })
 		},
 	},
 	methods: {
@@ -145,7 +145,7 @@ export default {
 		addTask: function() {
 			var task = { summary: this.newTaskName }
 
-			task.due = moment().add(7, 'days').startOf('day').format('YYYY-MM-DDTHH:mm:ss')
+			task.due = moment().startOf('day').format('YYYY-MM-DDTHH:mm:ss')
 			task.allDay = this.$store.state.settings.settings.allDay
 
 			this.createTask(task)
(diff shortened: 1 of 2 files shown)


jakobroehrl:
Hey @raimund-schluessler 
Thanks for implementing. It works great. :)

I thought about the due date. I work almost only with the week view and move the tasks around different dues. When I add a new task I want to see it on top of it. Then it's easy to move it to another date. And there are less task I want to do in 7 days. 
So I suggent so set the due date in the week view to today or tomorrow. I saw this in other todo tools (eg ticktick), too. 

What do you think?

raimund-schluessler:
I don't have a strong preference about which day we set the due date in week view. I just thought it makes sense to set it to +7 since it is *week* view. But I am fine to set it to today or tomorrow as well.

Maybe @skjnldsv or @jancborchardt have an opinion as well?

skjnldsv:
That's a valid question.
What do people expect in that case :thinking: 
i guess we can try 7 days and if people complain it will be easy to adjust :)

jakobroehrl:
> What do people expect in that case 🤔
> i guess we can try 7 days and if people complain it will be easy to adjust :)

I would complain...
What about a user default due setting?

raimund-schluessler:
> What about a user default due setting?

We definitly don't do settings for such values.

jancborchardt:
I would agree with @jakobroehrl’s point and just set it to today or tomorrow. :)

@jakobroehrl if you check further todo apps, maybe you can find out what is the most prevalent (today, or tomorrow) and we just use that then?

raimund-schluessler:
Then let's set it to today. Tomorrow seems a bit arbitrary to me. If others do it differently, then we can still adjust.

raimund-schluessler:
Should be good now.

Issue #778, opened 2019-12-23
Multiselect fires twice on select


raimund-schluessler:
After sharing with a new person or group the sharee occurs twice in the sharee list. After a page reload, only one occurence is there anymore.

raimund-schluessler:
This is an upstream issue of the vue-components. `@input` of the `multiselect` component is triggered twice for some reason. @skjnldsv The same happens for https://github.com/nextcloud/contacts when you use the latest `@nextcloud/vue`. I will report this upstream (and send a PR to contacts to migrate to `@nextcloud/vue` once I find the time).

raimund-schluessler:
Also every other multiselect now fires twice when an option is selected. This is especially annoying when the calendar is changed in the details view, because you can't move a task to a calendar twice and this always gives an error.

raimund-schluessler:
Fix is in https://github.com/nextcloud/nextcloud-vue/pull/844

PR #782, opened 2019-12-28
Only add sharee if not present already

raimund-schluessler:
This prevents to add a sharee twice.

### DIFF
--- a/src/components/CalendarShare.vue
+++ b/src/components/CalendarShare.vue
@@ -116,7 +116,8 @@ export default {
 			if (query.length > 0) {
 				const results = await client.principalPropertySearchByDisplayname(query)
 				this.usersOrGroups = results.reduce((list, result) => {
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
https://github.com/nextcloud/tasks/pull/782/commits/f2a3f000c21ff9c587aa5075f28bae3753fb19fc prevents to add a sharee to the list of sharees twice, https://github.com/nextcloud/tasks/pull/782/commits/facb0ab40e892a6a30169e55de228db9aff980b0 hides a user or group in the list of possible sharees if the resource is already shared with him.

Related to #778. 

@skjnldsv I think it makes sense to add this patch to contacts as well. Should I send a PR there?

skjnldsv:
> @skjnldsv I think it makes sense to add this patch to contacts as well. Should I send a PR there?

sure :)

PR #805, opened 2020-01-17
Fix tasks.svg (Qt QML: Unsupported image format)

misch7:


nickvergessen:
This is required for nice visual representation with https://github.com/nextcloud/desktop/pull/1565 inside the desktop client, so backporting would be nice @raimund-schluessler

raimund-schluessler:
Fine with me, @misch7 could you just sign-off your commit, please?

@nickvergessen Tasks currently supports Nextcloud down to 15, so the next version should cover everything.

nickvergessen:
The change is small enough to not qualify for signoff requirement

PR #818, opened 2020-01-22
Use the vue-components for app-navigation and app-content

### LINKED ISSUE #804: Drag and Drop on iOS 13.3 only selects text
Trying to drag and drop tasks on iOS 13.3 only selects text but does not actually drag anything.
I don't know when this broke, but maybe `user-select: none` might help.

raimund-schluessler:
This should fix #804 and lead to a more uniform look of the Tasks app.

Left to do for this PR:
- [x] Find a way to have drag and drop on collections and lists/calendars
- [x] Implement adding a task list
- [x] Properly style the list editing GUI
- [x] Implement the timeout GUI for deleting a calendar

While working on this, I noticed a few other things that should be improved in follow-up PRs:
- Enable to unshare a calendar #819 
- Use the colorpicker from the vue-components #820
- Show the user avatars for calendars shared with me and in the share component #821
- Switch to newly created calendar after creation #822

Another thing that should be fixed in the vue-components is the opacity of inactive navigation items and their hover state: https://github.com/nextcloud/nextcloud-vue/issues/425. Although the proposal https://github.com/nextcloud/nextcloud-vue/issues/425#issuecomment-527914136 looks very good, there wasn't much progress on implementing it lately. @skjnldsv Should we work-around it here or wait for the issue to be resolved?


css/src/style.scss
```diff
@@ -3,72 +3,44 @@
...
-				}
-			}
-		}
-	}
 	> ul {
 		> li {
-			background-image: unset;
-
-			> a {
+			&.collection .app-navigation-entry-icon {
```
skjnldsv: theoritecally you shouldn't need any custom css for your appnavigation :thinking:

raimund-schluessler: That's true, but I wanted to use the sprites file for the icons.

skjnldsv: Ah right, then I'm sure there is a lot of scss that can still go away ;)

raimund-schluessler: Definitly, already 100 lines SCSS less 🙂

raimund-schluessler:
This should be good now. Drag and drop also works and even sets the background color on hover (which it didn't before). There might be a problem with dropping a task on the app-navigation on touch devices, but I cannot check that since I don't own a touch device large enough for showing the app-navigation *and* the task list. However, if it really doesn't work there is an alternative way to dropping tasks on the sidebar.

raimund-schluessler:
Ping @skjnldsv, after #837 it's time to review this PR 😉

PR #843, opened 2020-01-29
Use eslint-config-nextcloud

### LINKED ISSUE #837: Use nextcloud eslint config
See https://github.com/nextcloud/eslint-plugin

raimund-schluessler:
With this PR the Tasks code now follows the `eslint-config-nextcloud` rules. Closes #837 

@skjnldsv This one is for you 😉


.eslintrc.js
```diff
@@ -1,11 +1,4 @@
 module.exports = {
-	root: true,
-	env: {
-		browser: true,
-		es6: true,
-		node: true,
-		jest: true
-	},
 	globals: {
 		t: false,
```
skjnldsv: is this needed?

src/components/Confirmation.vue
```diff
@@ -48,21 +48,20 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 import ClickOutside from 'vue-click-outside'
 
 export default {
-	name: 'PopoverMenu',
```
skjnldsv: No name?

.eslintrc.js
raimund-schluessler: Not sure, I will check.

src/components/TaskBody.vue
```diff
@@ -21,18 +21,16 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
 <template>
-	<li	v-show="showTask"
```
skjnldsv: I would never have caught that xD

src/components/Confirmation.vue
raimund-schluessler: I can add it back, but I think it is not necessary. The directive works and doesn't raise an error in the console. Also, this directive won't be used anymore after #818 is merged.

src/components/TaskBody.vue
raimund-schluessler: Me neither 😄

skjnldsv:
> Btw, with the new github actions, you can get linting comments in your pr directly 
> I can provide a pr

https://github.com/nextcloud/tasks/pull/844

Issue #821, opened 2020-01-22
Show the user avatars for calendars shared with me and in the share component


raimund-schluessler:
Related to #818

raimund-schluessler:
PR is #862.

Issue #819, opened 2020-01-22
Allow to unshare a calendar shared with me


raimund-schluessler:
Related to #818

raimund-schluessler:
PR is #862.

PR #862, opened 2020-02-10
Prevent editing not public tasks in lists shared with me

### LINKED ISSUE #467: Editing 'Show only busy' task always changes summary to 'busy'
I am actually not sure, whether this is a server issue, if we should handle this in the CalDAV apps or if this is a non-issue. Originally reported in https://github.com/nextcloud/tasks/pull/412#issuecomment-506845619. Input is very welcome.

### Steps to reproduce
1. Share a calendar with write privileges to user B.
2. Create a task in said calendar and set the privacy setting to `show only busy`.
3. User B edits the task.

### Actual behaviour
The edits are saved (as one would expect?), but the title is set to `Busy` if nothing else is put there.

### Expected behaviour
Actually not really sure, what should happen. User B has write privileges to the calendar, so I guess it is ok, that he can change tasks. But for a `show only busy` task he doesnt get any more information than the title `Busy`, so changing any property potentially overwrites values without him knowing, as it happens for the title for example.

**Tasks version:**
v0.11.0 (as this supports sharing only), but happens before if you share via Calendar app.

Since the same thing happens in the Calendar app, I would like to hear what @nextcloud/calendar thinks about this. Potentially also @nextcloud/designers has an opinion.

raimund-schluessler:
This PR fixes #467 and implements #819 and #821.

Editing tasks in a calendar shared with me which have an access class other than `PUBLIC` is not possible anymore. Also changing the access class for tasks in a calendar shared with me will be prevented.

However, this only fixes the problem on the client side. Editing such tasks still needs to be prevented on the server side as well, see https://github.com/nextcloud/server/issues/5551.

As a side effect, since we have to load the current user principal for this, we now show user avatars for the owner of a calendar shared with me in the app-navigation list and allow to unshare a calendar.

@rkaraba you might want to have a look as well.


raimund-schluessler:
@georgehrke How should we handle deleting non-public tasks/events in calendars shared with me? Do you plan to allow or forbid this in server? In this PR I treat such tasks as read-only, so deleting is not possible. But I would like to implement it the same way as in server.

raimund-schluessler:
@georgehrke Ping.

georgehrke:
@raimund-schluessler Yes, my plan for the server is to forbid editing / deleting.

PR #866, opened 2020-02-12
Move task-body actions into Action dropdown

### LINKED ISSUE #854: Alignment of the column
I suggest aligning task columns. Currently, if the task does not have a description, the task end date is shifted to the right in the list. This causes an unsightly appearance. In place of the details icon, you can leave the field blank or the icon inactive.

obraz

raimund-schluessler:
This PR moves the action icons / buttons into an Action component dropdown. This cleans up the task-body and allows to align the dates and icons better. We also have a delete action in the drop-down now which was requested as well at some point (and it makes sense, we have that for files as well).

It looks like this:
Screenshot_2020-03-02 Tasks - Nextcloud

With menu open:
Screenshot_2020-03-02 Tasks - Nextcloud(2)

The alignment of the date is not perfect, since for some tasks there are no actions to show (tasks in read-only calendars without subtasks). And since the due dates might have different widths the position of the icons left of the due dates might also shift. But I think it is as good as it gets in terms of alignment. Also, it looks a bit cleaner now, since most of the icons are now in the action dropdown.

Closes #854.

Feedback would be welcome @jancborchardt @skjnldsv @nextcloud/designers.


raimund-schluessler:
Updated my commen https://github.com/nextcloud/tasks/pull/866#issue-374563548. PR is ready to be merged.

Valdnet:
In my opinion, the date column should have a constant width. This would exclude shifting of the "Description" icon of the task, which for eg "Today" and "Yesterday" appears on the date field. It slightly disrupts the visualization.

obraz

obraz
Isn't that better? I know it takes up some space, but that's enough you can afford 😉.

raimund-schluessler:
> In my opinion, the date column should have a constant width. This would exclude shifting of the "Description" icon of the task, which for eg "Today" and "Yesterday" appears on the date field. It slightly disrupts the visualization.

I also thought about that, but the problem is that the width of the due date might vary with different languages. We would need to ellipsise the date in case this happens. I don't know if we want this.

skjnldsv:
> I also thought about that, but the problem is that the width of the due date might vary with different languages. We would need to ellipsise the date in case this happens. I don't know if we want this.

Yep, this is the issue without subgrid or table :(
I'd put it to a sufficient size and pray for the best :grin:

Valdnet:
@raimund-schluessler This takes the minimum table width for the date and the maximum width for the word. I think that in another language they are not long enough to take up a lot of space.

raimund-schluessler:
> I'd put it to a sufficient size and pray for the best 😁

Let's do it like this.

raimund-schluessler:
I set a fixed width for the date now and put an ellipsis in case it overflows. Looks quite good.

I realised the delete action might be clicked accidentally now that it is near the other actions, so I will add an undo option like we have it for the calendars list.

- [ ] Implement delete undo

raimund-schluessler:
Looks like this now:

Screenshot_2020-03-06 Tasks - Nextcloud(1)

Task scheduled for deletion:

Screenshot_2020-03-06 Tasks - Nextcloud

Valdnet:
@raimund-schluessler  It looks awesome now. Undoing task deletion - a very good idea :trophy:.

PR #965, opened 2020-04-18
Adjust Tasks to new nextcloud/vue version

raimund-schluessler:
This PR adjusts the app to nextcloud/vue@2.0.0.

tasks_vue

@skjnldsv @jancborchardt @ma12-co On small screens the main content is not shifted to the right anymore when the app-navigation is opened, so the app-navigation overlaps the content until you close it. Is this the new desired behaviour (I am ok with it, just asking to make sure nothing from Tasks is interfering with the new version)?

Edit: Just saw the answer in https://github.com/nextcloud/nextcloud-vue/pull/989:
>   * [x]  In mobile, opening the navigation does NOT shift the content, it just hovers it now. It avoid changing the whole layout and improve the rendering performances

But it might be nice to close the app-navigation on mobile when clicking outside of it, wouldn't it? Otherwise you have to manually close it again before you can access the content again.

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -2,7 +2,7 @@
 	* rules for app-navigation
 	*/
 
-#app-navigation {
+.app-navigation {
 	> ul {
 		> li {
 			&.collection .app-navigation-entry-icon::before {
@@ -241,15 +241,10 @@
 	text-decoration: none;
 }
 
-#content {
-	color: $gray_dark;
-}
-
-#app-content {
-	overflow-x: hidden;
-
+.app-content {
 	.header {
 		margin: 12px 0;
+		margin-left: 27px;
 		display: flex;
 
 		#add-task {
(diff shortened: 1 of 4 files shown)


skjnldsv:
> But it might be nice to close the app-navigation on mobile when clicking outside of it, wouldn't it? Otherwise you have to manually close it again before you can access the content again.

Would be cool, I think so too. :) @jancborchardt ?

jancborchardt:
> > But it might be nice to close the app-navigation on mobile when clicking outside of it, wouldn't it? Otherwise you have to manually close it again before you can access the content again.
> 
> Would be cool, I think so too. :) @jancborchardt ?

Yes – as the navigation overlaps the content now, this should happen. :) (Also need to make sure that the navigation closes on mobile when you trigger a content change with the navigation (but not for non-content-changing options like deleting an item, or opening the 3-dot menu).

raimund-schluessler:
I created an issue https://github.com/nextcloud/nextcloud-vue/issues/1070.

PR #992, opened 2020-04-29
Also show due time in task body on large screens

### LINKED ISSUE #914: [Feature Request] Show whole due (date and time) in the list
When a due is set the list shows only the date but not the time.
It would be useful if I can see date and time when I plan my day, now I have to open every appointment an keep in mind, what due time it has.

raimund-schluessler:
On large screens we also show the due time in the task body.

duetime

Closes #914.

@GambaJo @zeroepix Please have a look.

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -600,7 +600,29 @@
 									text-align: right;
 									text-overflow: ellipsis;
 									white-space: nowrap;
+								}
+
+								&__short {
 									width: 60px;
+								}
+
+								&__long {
+									width: 100px;
+									display: none;
+
+									&--date-only {
+										width: 60px;
+									}
+								}
+
+								@media only screen and (min-width: $breakpoint-mobile) {
+									&__long {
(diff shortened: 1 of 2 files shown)


zeroepix:
Looks good. I would certainly appreciate that for the days that I have tighter schedules to follow.

Does it make it extra crowded in the week view? That's the one that needs extra protection from too much info.

raimund-schluessler:
> Does it make it extra crowded in the week view? That's the one that needs extra protection from too much info.

I don't think it is a problem in the week view:
duetime_week

zeroepix:
Are you saying my tasks are too long? :)

image

raimund-schluessler:
> Are you saying my tasks are too long? :)

No 😄 This is a testing instance with only test names.
But even with longer names, I think there is enough space. And if a bit of the task name is cutoff, I guess you should still be able to recognize which task it is from the rest of the name 😉

Issue #1012, opened 2020-05-08
[Feature Request] Simple color change


javisomoza:
Hi

Sorry in case im doing something wrong; im not developer and im not used to using github and sending issues or a feature request.

This is a simple feature in which i have had to change only 2-3 lines of code:
![image](https://user-images.githubusercontent.com/49761255/81430710-2ad96680-9160-11ea-8ccb-f93da3e78e61.png)

I think that maintaining the list color in the tasks groups its a small change but enhances the look&feel considerably. 

Thanks for the app! Its great!

raimund-schluessler:
@javisomoza Welcome to Nextcloud 🙂 

I think it is a nice idea to show the calendar color in the task list. But showing it for every task creates to much visual noise for my taste and reduces the discoverability of the task which is currently opened in the right sidebar.

How about we show the calendar color next to the calendar name. This would also be consistent with how we show the calendar color in the app-navigation, right details sidebar and week view.
![Screenshot_2020-05-08 Tasks - Nextcloud](https://user-images.githubusercontent.com/2496460/81439860-5dd72680-916f-11ea-820e-824e59fff5c9.png)

PR is #1013.

javisomoza:
Thanks for your response, Raimund, and glad you are interested in the request.

Ill try to do my best explaining because i can do English grammatical mistakes jaja.
I like the idea of showing the color on each task to receive more info just the moment you look the task list and it also would allow (when showing "All" tasks together) to show all tasks mixed if you prefer (giving that option) and reduce opacity to reduce the noise. Anyway, each person works in his own way and has its preferences, even more when talking about colors and I completely understand what you say related to the visual noise.

Let me another small suggestion: the left border color in blue when hovering in a task could be the same as the calendar instead of blue?

Anyway, what you said about showing the calendar color next to the calendar name would be great.

Thanks!

raimund-schluessler:
> and it also would allow (when showing "All" tasks together) to show all tasks mixed if you prefer (giving that option)

Hm, I don't see a benefit of mixing all tasks in such a view. In the current "All" view you also see every task, but they are additionally sorted by calendar, which makes a lot of sense, I think.

> Let me another small suggestion: the left border color in blue when hovering in a task could be the same as the calendar instead of blue?

I would be ok to do that. But I think we will get rid of the left border sooner or later, since it is already not used anymore in most of the current apps (e.g. in the left sidebar). Is this correct @jancborchardt?

[...]

PR #1014, opened 2020-05-09
Adjust task-body background-color to new scheme

raimund-schluessler:
As discussed in https://github.com/nextcloud/tasks/issues/1012#issuecomment-626052255 this PR adjusts the task-body background-color on active and hover to comply with the new scheme e.g. used in nextcloud-vue.

Before (second task is hovered, first one active/opened):
Screenshot_2020-05-09 Tasks - Nextcloud(1)

After (second task is hovered, first one active/opened):
Screenshot_2020-05-09 Tasks - Nextcloud

@jancborchardt @javisomoza I would be open to replace the color used on active (`var(--color-primary-light)`) with the calendar color, but I have no idea how to get a 10% transparent color and apply it to the hover selector with javascript.

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -461,10 +461,12 @@
 						position: relative;
 						border: 1px solid $gray_light;
 
-						&:hover,
+						&:hover {
+							background-color: var(--color-background-hover);
+						}
+
 						&.active {
-							background-color: $gray_easy;
-							box-shadow: inset 4px 0 var(--color-primary);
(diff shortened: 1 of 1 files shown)


jancborchardt:
Very nice, already looks great! :)

> I would be open to replace the color used on active (`var(--color-primary-light)`) with the calendar color, but I have no idea how to get a 10% transparent color and apply it to the hover selector with javascript.

Any idea there @juliushaertl @skjnldsv @ma12-co?

PR #1125, opened 2020-07-15
Removing the import of the vars file

timkrief:
I could be wrong but it seems like the vars file have been removed from the project since the 190bb435fd3425bacb9f15b4e44e788bd18cc64f commit

### DIFF
--- a/css/tasks.scss
+++ b/css/tasks.scss
@@ -19,7 +19,6 @@
  *
  */
 
-@import './src/vars';
 @import './src/sprites-color';
 @import './src/sprites-bw';
 @import './src/style';


raimund-schluessler:
Good catch. If you plan on contributing more, we should make you a member of the Nextcloud organization, so you can create branches on the repo.

timkrief:
@raimund-schluessler sure! I'm working on the design PR right now so I might contribute more indeed 😉

raimund-schluessler:
@jancborchardt Could you please add @timkrief to the Tasks repo?

Issue #664, opened 2019-10-10
Update the design


jakobroehrl:
The sidebar looks a litte bit outdated. 
Maybe we could do some polish and use elements of the deck or calendar app (each from the vue branch):

Deck:
![grafik](https://user-images.githubusercontent.com/24530680/66557874-e8f92b00-eb52-11e9-9c3d-6c28865bed07.png)

---

Calendar:
![grafik](https://user-images.githubusercontent.com/24530680/66558377-c0bdfc00-eb53-11e9-8f45-ca9df158a042.png)

raimund-schluessler:
I also thought about updating the sidebar. Especially I would like to

-  introduce tabs since the amount of information in the sidebar increased over time
- add an illustration same as the Calendar app does it now

What I want to keep as it is, is that I want to show input elements only after the user clicks on something he wants to edit. I find all these input fields visually noisy.

jakobroehrl:
@raimund-schluessler 
I would like to help here, maybe after the Deck VUE migration is done.

timkrief:
On top of the sidebar, the task list was a bit intimidating for me.

I don't know if this could help, but I made a custom CSS style-sheet for my personal use. It's full of tricks right now since I'm only using a CSS on top of what's already existing, but if the general direction is of your liking, I'm willing to properly change the style of the app to make a PR.
![image](https://user-images.githubusercontent.com/14089949/87147429-f2afea80-c2ac-11ea-8797-1d2c074ce2b0.png)
(the background could be removed I just like bright colors!)
![image](https://user-images.githubusercontent.com/14089949/87147655-4ae6ec80-c2ad-11ea-99ea-c6940985170e.png)

raimund-schluessler:
I think it looks very nice, but it somehow doesn't match the overall nextcloud style. What do the @nextcloud/designers think?

timkrief:
Maybe the background could get in the way of judging the style, so here's with a neutral one: 
![image](https://user-images.githubusercontent.com/14089949/87151419-132f7300-c2b4-11ea-8b94-fd5f56600038.png)

I'm open to potential feedback and changes to try and fit to the overall nextcloud style :smiley:

juliushaertl:
I think that it would fit quite good if the shadows and border radius would just use the amount we have by default in our css variables, worth a try with that imo.

timkrief:
Ok, I used the CSS variables as advised and the draft looks like this now: 
[...]

PR #1135, opened 2020-07-21
Update to the design of the task list

timkrief:
Here's an updated style.scss that matches the designs I shown in issue #664

The task list looks like this for light mode
image
and like this for dark mode
image

### DIFF
--- a/css/src/style.scss
+++ b/css/src/style.scss
@@ -437,18 +437,19 @@ $priority_colors: ($red_overdue, $red_overdue, $red_overdue, $red_overdue, $yell
 			*/
 			
 			@mixin remove-last-squared-corners($sublevels-left: 8) {
-				&:empty {
+				&::before,
+				&>ol:empty::before {
 					display: none !important;
 				}
 				@if $sublevels-left > 0 {
-					&>.task-item:last-child>.subtasks-container>ol {
+					&>ol>.task-item:last-child>.subtasks-container{
 						@include remove-last-squared-corners($sublevels-left - 1);
 					}
 				}
 			}
 			
-			.grouped-tasks>ol,
-			.sortable-ghost>.subtasks-container>ol {
+			.grouped-tasks,
+			.sortable-ghost>.subtasks-container {
 				@include remove-last-squared-corners;
 			}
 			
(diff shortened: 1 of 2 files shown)


jancborchardt:
Looks very nice @timkrief, only 1 feedback:
The lines between 2 rows on the same level are rounded off on the left, which looks a bit off. I’d say it looks better if they are just straight?

Also you are now invited to the Nextcloud organization on Github, so in the future you don’t need your fork anymore and can just open branches in the main repository for better collaboration (don’t worry, the master branch is protected so nothing can break). :)

timkrief:
Thanks for the feedback @jancborchardt I fixed the lines in a new commit 7d5fb07d8091abb7cb32788e632fa7e11acf4dda . It looks way better now.
image

timkrief:
Wait I still have to fix an issue when subtasks are hidden. I thought the structure for a task with no subtasks and a tasks with hidden subtasks was the same but it is not.

raimund-schluessler:
> 
> 
> Wait I still have to fix an issue when subtasks are hidden. I thought the structure for a task with no subtasks and a tasks with hidden subtasks was the same but it is not.

Let me know if you need any help. I can have a look in the evening if necessary.

timkrief:
@raimund-schluessler I'm adding a `subtasksHidden` class to the `task-item`, when `!showSubtasks` in `TaskBody.vue`, I think that will be enough to be able to select it and stylize it with CSS without touching to the structure which makes sense as it is.

timkrief:
Ok, I added the class which let me stylize that element as needed, tasks with hidden subtasks now work. Let me know if there's anything else.

raimund-schluessler:
I will have a look in the evening. But since there are quite some changes here, you will have to signoff your commit(s), see https://github.com/nextcloud/tasks/pull/1135/checks?check_run_id=894171552.

timkrief:
Sorry for the hiccups, it's my first time contributing to nextcloud. I signed off the commits, removed some of my CSS I wasn't using anymore and fixed an edge-case i forgot of a task with the adding subtasks input displayed but subtasks hidden at the end of the tree.

raimund-schluessler:
> Sorry for the hiccups, it's my first time contributing to nextcloud. I signed off the commits, removed some of my CSS I wasn't using anymore and fixed an edge-case i forgot of a task with the adding subtasks input displayed but subtasks hidden at the end of the tree.

Don't worry, take your time. I won't have time to look at it before the evening anyway.

raimund-schluessler:
It looks nice already, but there are some visual issues:

- On Firefox with 100% zoom level, some lines separating the tasks don't get rendered properly, see here
grafik
I think the issue is, that you use the subtasks-container to kind of hide the round corners when necessary. We should try to calculate whether a task item should have round corners and set the CSS accordingly. That means:
   - If a root task is the first task, it needs round corners at the top.
   - If a root task is the last task, it needs round corners at the bottom.
   - If a task no matter the level has subtasks shown, it needs a round corner at the bottom left and none at the bottom right.
   - Subtasks never have round corners at the top, but might have round corners at the bottom, if the task itself and all parent tasks are the last in their container.

   I think this should be doable with a bit of CSS and Javascript (to detect whether subtasks are visible). And it should work no matter the level, I hope.
- The header of the task-details now merges with the new background-color, which looks weird.
- I would prefer the calculation of the checkbox color https://github.com/nextcloud/tasks/pull/1135/files#diff-88cbe2e916273c5c244516d0ec6b33ffR530 to be done in Vue as we e.g. do it here https://github.com/nextcloud/tasks/blob/master/src/components/TheDetails.vue#L598.
- You removed the background-color on hover and active for the task body. I think we should keep it, as it aligns with how the left sidebar and most lists in Nextcloud behave.

raimund-schluessler:
I took the liberty to push your branch to the Tasks repository and create a new PR in #1136 so we can better collaborate on the PR (it's a bit cumbersome to push to someone elses branch, even if its allowed).

PR #1186, opened 2020-08-11
Russian localization fix

sergey-dryabzhinsky:
Couldn't find project on transifex.
So here a little patch for `ru` language.
Some copy-paste typos.


raimund-schluessler:
Thanks for the fix, but the app *is* translated via Transifex, see https://www.transifex.com/nextcloud/nextcloud/translate/#de_DE/tasks (I only have access to the German language, hence the link to de_DE, you have to adjust it to your language).

Directly committing the translation files won't work, as they would be overwritten by nextcloud-bot on the next Transifex sync.

sergey-dryabzhinsky:
@raimund-schluessler Thanks for link. Commited fix into transifex.

PR #1213, opened 2020-09-05
Fix fluttering sortorder popover

raimund-schluessler:
This fixes the fluttering of the sortorder dropdown when scrolling.


raimund-schluessler:
With #1224 the `boundaries-element` changes can be removed.

Issue #1251, opened 2020-10-03
Tasks doesn't work after upgrade Nextcloud from 19 to 20


nikorn:
### Steps to reproduce
1. Start a browser
2. Login to Nextcloud
3. Start Tasks

### Expected behaviour
Tasks page view inside Nextcloud environment

### Actual behaviour
Nextcloud crashes with error message:
Page unavailable. The site did not send data.
ERR_EMPTY_RESPONSE

Another Nextcloud applications works fine

### Server configuration

**Operating system**: Ubuntu Server 20.4

**Web server:** Apache 2.4.41

**Database:** MySQL 10.3.22

**PHP version:** 7.4

**Nextcloud version:** 20.0.0

**Tasks version:** 0.13.3

**Updated from an older Nextcloud or fresh install:** Nextcloud updated from 19.0 то 20.0

**Signing status:**

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
  - appointments: 1.7.7
  - apporder: 0.11.0
  - bruteforcesettings: 2.0.1
  - calendar: 2.1.2
  - cloud_federation_api: 1.3.0
  - cms_pico: 1.0.10
  - comments: 1.10.0
  - contacts: 3.4.0
  - contactsinteraction: 1.1.0
  - dav: 1.16.0
  - event_update_notification: 1.1.0
  - external: 3.7.0
  - federatedfilesharing: 1.10.1
  - files: 1.15.0
  - files_external: 1.11.1
  - files_markdown: 2.3.1
  - files_readmemd: 1.1.4
  - files_rightclick: 0.17.0
  - files_sharing: 1.12.0
  - files_trashbin: 1.10.1
  - files_versions: 1.13.0
  - files_videoplayer: 1.9.0
  - logreader: 2.5.0
  - lookup_server_connector: 1.8.0
  - nextcloud_announcements: 1.9.0
  - notifications: 2.8.0
  - oauth2: 1.8.0
  - ojsxc: 4.1.1
  - password_policy: 1.10.1
  - photos: 1.2.0
  - privacy: 1.4.0
  - provisioning_api: 1.10.0
  - qownnotesapi: 20.9.0
  - serverinfo: 1.10.0
  - settings: 1.2.0
  - tasks: 0.13.3
  - theming: 1.11.0
  - twofactor_backupcodes: 1.9.0
  - updatenotification: 1.10.0
  - user_status: 1.0.0
  - viewer: 1.4.0
[...]

PR #1261, opened 2020-10-10
[FIX] added a few variables that was causing SCSS issues in NC19.0.4+

LouisVallat:
This pull requests is aimed towards fixing the issue (https://github.com/nextcloud/tasks/issues/1251) some of us had with Tasks on the latest Nextcloud versions (NC19.0.4+).

As mentioned by @Terrox in the discussion thread, adding these variables fixes the SCSS compilation problem we had, that broke the Tasks App's CSS.

I tested this modification on a fresh vanilla docker version of NC19.0.3 as well as on my NC20 version, and everything looked all good.

### DIFF
--- a/css/src/markdown.scss
+++ b/css/src/markdown.scss
@@ -1,6 +1,11 @@
 /**
  * Markdown rendering
  */
+
+$color-lightgrey: #bbbbbb;
+$color-main-background: #ddd;
+$color-darkgrey:#666;
+
 #markdown {
 	width: 100% !important;
 	min-height: 40px;


paulcalabro:
@LouisVallat I think you might only want to add the variables that aren't defined [here](https://github.com/nextcloud/server/blob/master/core/css/css-variables.scss):

- **$color-lightgrey**

- **$color-darkgrey**

css/src/markdown.scss
```diff
@@ -1,6 +1,11 @@
 /**
  * Markdown rendering
  */
+
+$color-lightgrey: #bbbbbb;
+$color-main-background: #ddd;
```
paulcalabro, line 6: See comments, but I don't think line is needed since it's defined here:

https://github.com/nextcloud/server/blob/master/core/css/css-variables.scss

paulcalabro (changes requested):

raimund-schluessler:
Thanks for the debugging and the fix!

To be honest, I don't know why this ever worked, and just broke right now. I couldn't find `$color-lightgrey` nor `$color-darkgrey` in the server variables (present or history). Probably the SCSS compiler got more strict with a recent update and fails instead of warns now. 🙈

However, the SCSS color variables are anyway deprecated, see https://github.com/nextcloud/server/blob/master/core/css/variables.scss#L33. So we should move to CSS4 variables right away. I will provide a PR.

LouisVallat:
To be honest, I just did a pull request based on the fix someone provided on the issue thread.
I don't know either! Maybe the compiler replaced these values by some default ones, but didn't in the updated version provided with the NC server.
Good! I'm closing my pull request then! Can't wait to get that update so we can close the issues related to this bug on this repo as well and the ones opened on the server's one!
I'm glad I could help, that's the first pull request I did, and I would like to contribute more often to open source projects such as this one!

Have a nice day :smile:

PR #1285, opened 2020-10-13
Fix sorting calendars on mobile

### LINKED ISSUE #1278: Fix sorting calendars on mobile
Similar to how we sort tasks, sorting calendars on mobile should only start after a long click, not immediately.

raimund-schluessler:
Sorting calendars on mobile now only starts after a short delay, ensuring it is still possible to just scroll the list and click the calendar conveniently. Closes #1278.


raimund-schluessler:
Works well on iOS as long as you use the `icon-bullet` as drag handle. Otherwise Safari tries to preview the link, consequently preventing dragging the calendar item. I think this is good enough.

PR #1290, opened 2020-10-14
Fix calendar delete wording

### LINKED ISSUE #1279: "Deleting the calendar in {countdown}"
### Steps to reproduce
1. Delete a list on Tasks
2. Mouse over the 'revert'/'undo icon

### Expected behaviour
Correct language used

### Actual behaviour
image
Mentions a 'calendar', this is really confusing for users as they think they're deleting a calendar rather than a list, and ultimately panic, and undo the deletion.

raimund-schluessler:
- Don't show calendar delete option if the calendar is read-only (e.g. a list from Deck)
- Call calendars that don't support tasks "lists" to prevent confusion

Closes #1279.

### DIFF
--- a/src/components/AppNavigation/ListItemCalendar.vue
+++ b/src/components/AppNavigation/ListItemCalendar.vue
@@ -205,14 +205,26 @@ export default {
 		},
 
 		deleteMessage() {
-			return !this.calendar.isSharedWithMe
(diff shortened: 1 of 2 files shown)


raimund-schluessler:
I will wait for a couple of days before releasing a new version to let the translations being updated.

PR #1361, opened 2020-12-01
Add aCalendar+ to app list in README

dkerkow:
Using it myself and it works pretty good. Not affiliated.


raimund-schluessler:
Fine with me (but not tested) 👍

PR #1394, opened 2021-01-18
Fix permissions of markdown.scss 755 -> 644

beardhatcode:
For some reason some files are needlessly marked as executable in this repository. This fixes some of that.

Non 644 permissions in the tar:
```
-rw-r-xr-- ./img/sprt-bw.svg
-rw-r-xr-- ./img/sprt-color.svg
-rwxr-xr-x ./css/src/markdown.scss
-rwxr--r-- ./css/src/Calendars/Calendar.scss
-rw-r-xr-- ./js/tasks.js.LICENSE.txt
```

After these patches:
```
-rwxr-xr-x ./css/src/markdown.scss
-rwxr--r-- ./css/src/Calendars/Calendar.scss
-rw-r-xr-- ./js/tasks.js.LICENSE.txt
```

I don't immediately see how to fix the last 3.


raimund-schluessler:
@skjnldsv Do you know if this is necessary? Doesn't the server take care of the file permissions when installing an app from the appstore?
I am not so keen on having a `chmod` on the `Makefile`, to be honest.

skjnldsv:
> @skjnldsv Do you know if this is necessary? Doesn't the server take care of the file permissions when installing an app from the appstore?
> I am not so keen on having a `chmod` on the `Makefile`, to be honest.

No we'll keep permissions iirc.
But you should most likely not be this either. I never had an issue with that since i use http git lol

beardhatcode:
> Doesn't the server take care of the file permissions when installing an app from the appstore?

The server allows executable files to be included in a package (see camerarawpreviews). 

My build system (nix) was complaining about "suspicious permissions" on some files.

So that is the reason for this PR. To me it seems that we would want a minimum of executable files in the published artefacts. 


>  I am not so keen on having a chmod on the Makefile, to be honest.

I understand that you do not want to add the `chmod` to the makefile, this should actualy be fixed in the upstream, and I'll look into that when I have time. 

But could you consider only the changes purposed in https://github.com/nextcloud/tasks/pull/1394/commits/ec53098307b489011b31e9cda39e8e6923893621 (no code change only permission change) ?

beardhatcode:
The PR has been adjusted to only include the first commit

PR #1549, opened 2021-05-02
Don't scroll datepicker in AppSidebar with body

### LINKED ISSUE #1530: Don't scroll datepicker with body
Due to the way we fixed https://github.com/nextcloud/tasks/issues/935 in https://github.com/nextcloud/nextcloud-vue/pull/1880 the datepicker is not cut-off anymore, but now scrolls with body:

datepicker

In case we decide for the better solution in https://github.com/nextcloud/nextcloud-vue/pull/1881 we need to adjust the Tasks app accordingly.

raimund-schluessler:
This prevents scrolling the datepicker with body. Requires https://github.com/nextcloud/nextcloud-vue/pull/1881 and closes #1530.


raimund-schluessler:
Will likely not happen.

Issue #1552, opened 2021-05-03
Questions regarding grammar


rakekniven:
There is some wording that I don't understand.

How to create a task "in tomorrow"?
https://github.com/nextcloud/tasks/blob/3a44205ab7d621b106d3f5fbe12634e3eea0a37f/src/views/AppSidebar.vue#L339

How can I complete tasks for the next day? Do work in advance?
https://github.com/nextcloud/tasks/blob/3a44205ab7d621b106d3f5fbe12634e3eea0a37f/src/views/AppSidebar.vue#L319

How to modify a task in the future?
https://github.com/nextcloud/tasks/blob/3a44205ab7d621b106d3f5fbe12634e3eea0a37f/src/views/AppSidebar.vue#L329

Would be great to have some comments in source code for translators.

Where can I look up the definition for "L" or "LT"?

raimund-schluessler:
The strings `L` and `LT` are from moment.js: https://momentjs.com/docs/#/parsing/string-format/
Taken from the moment.js doc:
![Screenshot_2021-05-03 Moment js Docs](https://user-images.githubusercontent.com/2496460/116852038-e4440e00-abf3-11eb-8a11-8bbb431fab61.png)

Having created, modified and completed dates from the future, doesn't exactly make sense probably. But since we don't limit the dates to always be in the past, and other clients can edit these dates too, plus, we will allow to edit modified and completed dates manually (see https://github.com/nextcloud/tasks/issues/1184) I wanted to account for future dates as well.

rakekniven:
Thank you for your quick response.
Technical side I understand your description.
But I think the translators are lost seing three mentioned strings.

What about removing this past phrases together with "tomorrow" and using pure dates?
E. g. use "[Created at] LT" instead of "[Created tomorrow at] LT".

raimund-schluessler:
> What about removing this past phrases together with "tomorrow" and using pure dates?
> E. g. use "[Created at] LT" instead of "[Created tomorrow at] LT".

Since it's an edge case, I am fine with adjusting the future tense. But what is wrong with the past tense?

rakekniven:
> But what is wrong with the past tense?

Nothing, this is a misunderstanding. I was talking about "Created tomorrow" where "Created" is past.

PR #1558, opened 2021-05-04
Adjust wording for future dates

raimund-schluessler:
Closes https://github.com/nextcloud/tasks/issues/1552

### DIFF
--- a/src/views/AppSidebar.vue
+++ b/src/views/AppSidebar.vue
@@ -316,7 +316,7 @@ export default {
 				return this.task.completedDateMoment.calendar(null, {
 					lastDay: this.$t('tasks', '[Completed yesterday at] LT'),
 					sameDay: this.$t('tasks', '[Completed today at] LT'),
-					nextDay: this.$t('tasks', '[Completed tomorrow at] LT'),
+					nextDay: this.$t('tasks', '[Completed] L'),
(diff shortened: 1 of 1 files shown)


src/views/AppSidebar.vue
```diff
@@ -336,7 +336,7 @@ export default {
 				return this.task.createdMoment.calendar(null, {
 					lastDay: this.$t('tasks', '[Created yesterday at] LT'),
 					sameDay: this.$t('tasks', '[Created today at] LT'),
-					nextDay: this.$t('tasks', '[Created tomorrow at] LT'),
+					nextDay: this.$t('tasks', '[Created] L'),
 					lastWeek: this.$t('tasks', '[Created last] dddd [at] LT'),
 					nextWeek: this.$t('tasks', '[Created] dddd [at] LT'),
```
rakekniven, line 341: More comments arrive at Transifex regarding future completeness.

Can we use a more simple way for "nextWeek" like '[Created] L' as well?

raimund-schluessler: Fine with me, but `[Created] dddd [at] LT` is not really specific to a certain tense, isn't it? It just write out as `Created Monday at 12:34:56`.

rakekniven (commented):
You are right

PR #1562, opened 2021-05-05
l10n: Delete space

Valdnet:
Signed-off-by: Valdnet <47037905+Valdnet@users.noreply.github.com>

### DIFF
--- a/src/components/TaskBody.vue
+++ b/src/components/TaskBody.vue
@@ -67,7 +67,7 @@ License along with this library.  If not, see <http://www.gnu.org/licenses/>.
 				<div v-if="task.complete > 0" class="percentbar">
 					<div :style="{ width: task.complete + '%', 'background-color': task.calendar.color }"
 						:class="{'completed': task.completed}"
-						:aria-label="$t('tasks', '{complete} % completed', {complete: task.complete})"
+						:aria-label="$t('tasks', '{complete}% completed', {complete: task.complete})"
 						class="percentbar__done" />
 				</div>
 			</div>
--- a/src/views/AppSidebar.vue
+++ b/src/views/AppSidebar.vue
@@ -624,7 +624,7 @@ export default {
 			return null
 		},
 		completeString() {
-			return this.$t('tasks', '{percent} % completed', { percent: this.task.complete })
+			return this.$t('tasks', '{percent}% completed', { percent: this.task.complete })
 		},
 		targetCalendars() {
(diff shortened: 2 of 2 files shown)


raimund-schluessler:
@Valdnet The space is correct and I would like to keep it. Maybe due to my background, I really like SI units. And they state:

> The value of a quantity is written as a number followed by a space (representing a multiplication sign) and a unit symbol; e.g., 2.21 kg, 7.3×102 m2, 22 K. This rule explicitly includes the percent sign (%) and the symbol for degrees Celsius (°C).

(Taken from Wikipedia https://en.wikipedia.org/wiki/International_System_of_Units#General_rules)

Valdnet:
@raimund-schluessler According to the guide: [Style guide english](https://ec.europa.eu/info/sites/default/files/styleguide_english_dgt_en.pdf) the percentage sign of a number should be written right after the number, i.e.: 100%. Well, unless we use the [ISO standard](http://publications.europa.eu/code/en/en-360400.htm). This is also the case for text strings in other Nextcloud applications. We should standardize the spelling in Nextcloud.

raimund-schluessler:
> 
> 
> @raimund-schluessler According to the guide: [Style guide english](https://ec.europa.eu/info/sites/default/files/styleguide_english_dgt_en.pdf) the percentage sign of a number should be written right after the number, i.e.: 100%. Well, unless we use the [ISO standard](http://publications.europa.eu/code/en/en-360400.htm). This is also the case for text strings in other Nextcloud applications. We should standardize the spelling in Nextcloud.

In the style guide from the European Commission you linked, they also state that a (hard) space is present between value and unit (in accordance to the ISO standard) for official publications:

> Percentages. Note  that per  centis  normally  written  as  two  words  in  British English.  Use per  centwhere  the  number  is  also  spelled  out  in  words: twenty per  cent.  With  figures,  use  the  per  cent  sign  (%),  closed  up  to  the  figure,e.g. 25%. However,  in  official  (legal  and  non-legal)  publications,  a  hard  space1is added  automatically  before  the  percentage  symbol  in  accordance  with  the relevant  ISO standard. See  also Section6.4of  the Interinstitutional Style Guide.

So, as I read it, not having a space is simply wrong. And I don't really understand why they sometimes don't do it. I agree we should unify that throughout Nextcloud, but towards the correct solution with a space.

And, ideally, this should be a hard space (no line-break allowed), but this will get lost in translation, so I am fine with a simple space.

PR #1579, opened 2021-05-10
Use text-box-outline for notes icon

### LINKED ISSUE #1577: use text-box-outline instead of NoteText?
Hi @raimund-schluessler, 
I think the text-box-outline (might be called differently) icon does fit better for notes than the NoteText icon. What do you think? 
I've created a fast mockup:
| Before | After |
|---|---|
| image | image |

(of course this should then also be changed in the list view)

raimund-schluessler:
Screenshot_2021-05-10 Tasks - Nextcloud(1)

Closes #1577.


szaimen:
Thanks!

PR #1594, opened 2021-05-15
Use codecov action

### LINKED ISSUE #1440: Don't use codedov secrets anymore
We should stop using codecov tokens as they are not needed for public repositories. Needs to be changed at these positions:
https://github.com/nextcloud/tasks/blob/e344ef4d2a361b16e4a467f6d009d619209601bd/.github/workflows/phpunit.yml#L83
https://github.com/nextcloud/tasks/blob/b3b8434fd24ec2811b8c294174bdfe5e47f5363f/.github/workflows/node-test.yml#L37

See here for how to:
https://github.com/nextcloud/end_to_end_encryption/pull/227/files#diff-ea12f60188cdd90bc99e5d0af2eb91647bbe4a9199176aa1ec5240f65efed510L217

@tobiasKaminsky For reference.

raimund-schluessler:
Closes #1440.

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -27,11 +27,14 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Set up Codecov
-        run: npm install -g codecov
       - name: Test
         run: npm run test
         env:
           CI: true
-      - name: Upload coverage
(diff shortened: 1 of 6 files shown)


raimund-schluessler:
@tobiasKaminsky Somehow the coverage of the PHP files is not calculated. Do you have an idea why?

PR #1638, opened 2021-06-11
Use npm v7 and simplify overall configuration

raimund-schluessler:
This PR moves the app to npm v7 and simplifies the overall configuration. See https://github.com/nextcloud/app-tutorial/pull/319.


skjnldsv:
We didn't really agreed to move to  :p
That's why there was this pr https://github.com/nextcloud/app-tutorial/pull/319 and  we asked on the frontend channel :grin: 

So far it seems to work, but I don't know about others apps/libs :see_no_evil:

raimund-schluessler:
> We didn't really agreed to move to :p

Hm, seems I was a bit to fast with that one. 🙈 But since it works quite well, I don't really want to go back. And someone has to do the first step 😄

skjnldsv:
The issue is that either everyone move to that or no one. Switching versions back and forth is not acceptable :(
So if some apps breaks, we're kind of blocked :'(

package.json
```diff
@@ -60,51 +60,19 @@
...
 		"@babel/plugin-proposal-object-rest-spread": "^7.14.5",
-		"@babel/plugin-syntax-dynamic-import": "^7.8.3",
-		"@babel/preset-env": "^7.14.5",
+		"@nextcloud/babel-config": "^1.0.0-beta.1",
 		"@nextcloud/browserslist-config": "^2.1.0",
 		"@nextcloud/eslint-config": "^5.1.0",
-		"@nextcloud/eslint-plugin": "^2.0.0",
+		"@nextcloud/stylelint-config": "^1.0.0-beta.0",
 		"@nextcloud/webpack-vue-config": "^4.0.3",
 		"babel-core": "^7.0.0-bridge.0",
```
skjnldsv, line 69: Should not be needed

skjnldsv: Ah, for jest?

package.json
```diff
@@ -60,51 +60,19 @@
 		"vuex-router-sync": "^5.0.0"
 	},
 	"devDependencies": {
-		"@babel/core": "^7.14.5",
-		"@babel/eslint-parser": "7.14.5",
 		"@babel/plugin-proposal-object-rest-spread": "^7.14.5",
```
skjnldsv, line 63: Already within `@babel/preset-env`

skjnldsv: https://babeljs.io/docs/en/babel-plugin-proposal-object-rest-spread

package.json
raimund-schluessler: Yes, this is for jest. Although I didn't test it without.

package.json
raimund-schluessler: > 
> 
> The issue is that either everyone move to that or no one. Switching versions back and forth is not acceptable :(
> So if some apps breaks, we're kind of blocked :'(

Yes, I know. But I guess, we will only know if something breaks if everyone gives it a try. I am fine to move back to npm6 as an interim solution in case something really breaks. However, we won't be able to stay with npm6 forever anyway.

raimund-schluessler:
> 
> 
> The issue is that either everyone move to that or no one. Switching versions back and forth is not acceptable :(
> So if some apps breaks, we're kind of blocked :'(

Yes, I know. But I guess, we will only know if something breaks if everyone gives it a try. I am fine to move back to npm6 as an interim solution in case something really breaks. However, we won't be able to stay with npm6 forever anyway.

package.json
raimund-schluessler: https://github.com/nextcloud/tasks/pull/1640

PR #1649, opened 2021-06-23
Create task from Talk message

jakobroehrl:
Signed-off-by: Jakob Röhrl <jakob.roehrl@web.de>

**Todo**
- [ ] Style layout https://github.com/nextcloud/tasks/issues/1682
- [ ] Select the default task list instead of the first one https://github.com/nextcloud/tasks/issues/1687

---

**Screenshots:**
Screenshot 2021-06-24 at 11-02-54 Talk updates ✅ - Talk - Nextcloud

image

---
references:
https://github.com/nextcloud/deck/pull/2795

### DIFF
--- a/webpack.js
+++ b/webpack.js
@@ -4,6 +4,7 @@ const path = require('path')
 webpackConfig.entry = {
 	...webpackConfig.entry,
 	dashboard: path.join(__dirname, 'src', 'dashboard.js'),
+	talk: path.join(__dirname, 'src', 'talk.js'),
 }
 
 module.exports = webpackConfig
--- a/css/tasks-talk.scss
+++ b/css/tasks-talk.scss
@@ -1,0 +1,1 @@
+@include icon-black-white('tasks', 'tasks', 1);
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -23,10 +23,12 @@
 namespace OCA\Tasks\AppInfo;
 
 use OCA\Tasks\Dashboard\TasksWidget;
+use OCA\Tasks\Listeners\BeforeTemplateRenderedListener;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
+use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
 
 class Application extends App implements IBootstrap {
 
@@ -42,6 +44,8 @@ class Application extends App implements IBootstrap {
 
 	public function register(IRegistrationContext $context): void {
 		$context->registerDashboardWidget(TasksWidget::class);
+
+		$context->registerEventListener(BeforeTemplateRenderedEvent::class, BeforeTemplateRenderedListener::class);
 	}
 
 	public function boot(IBootContext $context): void {
--- a/lib/Listeners/BeforeTemplateRenderedListener.php
+++ b/lib/Listeners/BeforeTemplateRenderedListener.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ * @copyright Copyright (c) 2021 Jakob Röhrl <jakob.roehrl@web.de>
+ *
+ * @author Julius Härtl <jus@bitgrid.net>
+ * @author Jakob Röhrl <jakob.roehrl@web.de>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+declare(strict_types=1);
+
+
+namespace OCA\Tasks\Listeners;
+
+use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\IRequest;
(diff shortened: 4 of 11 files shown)


webpack.js
```diff
@@ -1,3 +1,10 @@
 const webpackConfig = require('@nextcloud/webpack-vue-config')
 
+const path = require('path')
+
+webpackConfig.entry = {
+	...webpackConfig.entry,
+	dashboard: path.join(__dirname, 'src', 'init-talk.js'),
```
raimund-schluessler, line 7:
```suggestion
	talk: path.join(__dirname, 'src', 'init-talk.js'),
```

jakobroehrl: I don't see the action after changing this, do you have another idea?

raimund-schluessler:
@jakobroehrl The problem was that you never registered the `BeforeTemplateRenderedListener`, so you the Javascript file was never actually loaded. I fixed that, so now the file is loaded and an entry `Create new task` shows up in the Talk message action:

Screenshot 2021-06-24 at 11-02-54 Talk updates ✅ - Talk - Nextcloud

When you click it, the modal opens and throws a lot of errors. This is tbd yet.

jakobroehrl:
Thanks a lot! :)
I work on the create dialog and maybe we can use it for the dashboard to add an task directly

jakobroehrl:
@raimund-schluessler 
How to get all the calendars from the actual user into this file? Thanks!
https://github.com/nextcloud/tasks/blob/0ed0c6d98118622f32d2da1d9b5b99fcc9c11b9a/src/components/TaskCreateDialog.vue

Played a little bit around but couldn't figure it out.

raimund-schluessler:
> 
> 
> @raimund-schluessler
> How to get all the calendars from the actual user into this file? Thanks!
> https://github.com/nextcloud/tasks/blob/0ed0c6d98118622f32d2da1d9b5b99fcc9c11b9a/src/components/TaskCreateDialog.vue

Similar to how you get them here:
https://github.com/nextcloud/tasks/blob/ed10468daf82b1cf4792f183da53b3ef62d4ddbf/src/views/AppSidebar.vue#L646-L647

But you also need to request them first, of course.
https://github.com/nextcloud/tasks/blob/ed10468daf82b1cf4792f183da53b3ef62d4ddbf/src/App.vue#L59-L60

jakobroehrl:
@raimund-schluessler 
this.$store is undefined 
Tried to import it but it didn't work out.

raimund-schluessler:
> 
> 
> @raimund-schluessler
> this.$store is undefined
> Tried to import it but it didn't work out.

Without seeing the whole code, it's difficult to know what's wrong. But I guess you didn't use `Vuex`. You can have a look at how it works for the dashboard:
https://github.com/nextcloud/tasks/blob/6a4dd9af562075881379062666ff7cb6e1c060f0/src/dashboard.js#L33-L35

raimund-schluessler:
@jakobroehrl I fixed the error you were seeing after loading the calendars. The problem is that you cannot simply render an array of Objects with `{{ objectsArray }}` in your template. The calendars are now loaded and shown correctly.

Screenshot 2021-07-01 at 22-20-12 Talk updates ✅ - Talk - Nextcloud

There is still quite a bit to do:

- [ ] Creating a task is non-functional
- [ ] Task summary and description should be set from the talk message content
- [ ] Layout needs style fixes
- [ ] Code needs a cleanup
- [ ] Optional: Select the default task list instead of the first one

raimund-schluessler:
>    * [ ]  Select the default task list instead of the first one: I always get the personal calendar here: https://github.com/nextcloud/tasks/blob/7bfe59e0dca63633b3f21a7ab6cef13099f8ac59/src/components/TaskCreateDialog.vue#L116-L120
>        Why?

The default calendar is not loaded here. Normally it comes from the initialState service, see https://github.com/nextcloud/tasks/blob/master/lib/Controller/PageController.php#L62
But here in Talk, this does not work, since the initial state is never loaded. So we need a separate endpoint in PHP to get only the default calendar and then query it in Javascript via axios.

jakobroehrl:
> > * [ ]   Select the default task list instead of the first one: I always get the personal calendar here: https://github.com/nextcloud/tasks/blob/7bfe59e0dca63633b3f21a7ab6cef13099f8ac59/src/components/TaskCreateDialog.vue#L116-L120
> >   
> >   Why?
> 
> The default calendar is not loaded here. Normally it comes from the initialState service, see https://github.com/nextcloud/tasks/blob/master/lib/Controller/PageController.php#L62
> But here in Talk, this does not work, since the initial state is never loaded. So we need a separate endpoint in PHP to get only the default calendar and then query it in Javascript via axios.

OK, I have an idea how this could be solved. I would do it in another PR, OK?

Could you maybe look over this: 
I don't get an result here back: https://github.com/nextcloud/tasks/blob/7bfe59e0dca63633b3f21a7ab6cef13099f8ac59/src/components/TaskCreateDialog.vue#L167

raimund-schluessler:
> Could you maybe look over this:
> I don't get an result here back:
> 
> https://github.com/nextcloud/tasks/blob/7bfe59e0dca63633b3f21a7ab6cef13099f8ac59/src/components/TaskCreateDialog.vue#L167

Looking at the `createTask` action in the store, I think it simply does not return anything. What do you need the Task for?

jakobroehrl:
> > Could you maybe look over this:
> > I don't get an result here back:
> > https://github.com/nextcloud/tasks/blob/7bfe59e0dca63633b3f21a7ab6cef13099f8ac59/src/components/TaskCreateDialog.vue#L167
> 
> Looking at the `createTask` action in the store, I think it simply does not return anything. What do you need the Task for?

I want to give a feedback to the user (or close the create modal) if the task was created.

raimund-schluessler:
> 
> 
> > > Could you maybe look over this:
> > > I don't get an result here back:
> > > https://github.com/nextcloud/tasks/blob/7bfe59e0dca63633b3f21a7ab6cef13099f8ac59/src/components/TaskCreateDialog.vue#L167
> > 
> > 
> > Looking at the `createTask` action in the store, I think it simply does not return anything. What do you need the Task for?
> 
> I want to give a feedback to the user (or close the create modal) if the task was created.

I guess you need to adjust the `createTask` action then.

raimund-schluessler:
@jakobroehrl Please have a look. I think it works ok, but needs some style fixes.

jakobroehrl:
> @jakobroehrl Please have a look. I think it works ok, but needs some style fixes.

It's working great! The looks a bit weired, but I didn't figure it out how to solve.
image

raimund-schluessler:
I will merge this now as it is, since it works as it should. Style fixes will come with https://github.com/nextcloud/tasks/issues/1682 later.

PR #1650, opened 2021-06-24
l10n: Change to a capital letter

Valdnet:


rakekniven (approved):

raimund-schluessler:
Sorry, but this needs to stay a small letter. This text is to be combined with the word `More` to then read `More upcoming tasks ... `, see here: https://github.com/nextcloud/nextcloud-vue-dashboard/blob/master/src/DashboardWidget.vue#L75 In this string `what` is supposed to be replaced by `upcoming tasks`.

I also find it a bit unfortunate, but that's how it is at the moment.

Valdnet:
@raimund-schluessler  Thanks for the clarification 😉. I was also wondering about the code `:show-more-text`.

> This text is to be combined with the word More to then read More upcoming tasks ...

It is a pity that this note is not included in the developer notes in Transifex for a text string. It would help in translating, because in this case the meaning of the sentence changes a bit.

rakekniven:
We definitely need comments for translators.

PR #1662, opened 2021-06-28
Remove scrutinizer config

raimund-schluessler:


raimund-schluessler:
I would like to get rid of the scrutinizer check, since it doesn't seem to be used in any other app anymore. However, the action is still there after removing the scrutinizer config file. Does anyone know if it is enough to remove the Webhook? Or does something else need to be done?

Maybe @skjnldsv or @MorrisJobke know?

skjnldsv:
Done

skjnldsv (approved):

Issue #1685, opened 2021-07-13
Moving task between lists back and forth causes error


carhe:
### Steps to reproduce
1. Create a task in some list A.
2. Move the task to another list B.
3. Move the task back to list A.

### Expected behaviour
Task should be moved successfully from A to B and back to A.

### Actual behaviour
Movement from A to B works, but trying to move back to A fails with a generic error message ("Es ist ein Fehler aufgetreten.").

This behavior appeared after the update from Nextcloud 21 to 22 with Tasks version 0.13.6 (it worked before the update). It did not change with Tasks version 0.14.0-alpha.1.

The Nextcloud log contains hints (see below).

### Server configuration

**Operating system**: Linux 5.4.0-77-generic #86-Ubuntu SMP Thu Jun 17 02:35:03 UTC 2021 x86_64

**Web server:** Apache/2.4.38 (Debian) (apache2handler)

**Database:** mysql 10.5.11

**PHP version:** 7.4.21
Modules loaded: Core, date, libxml, openssl, pcre, sqlite3, zlib, ctype, curl, dom, fileinfo, filter, ftp, hash, iconv, json, mbstring, SPL, PDO, session, posix, Reflection, standard, SimpleXML, pdo_sqlite, Phar, tokenizer, xml, xmlreader, xmlwriter, mysqlnd, apache2handler, apcu, bcmath, exif, gd, gmp, imagick, intl, ldap, memcached, pcntl, pdo_mysql, pdo_pgsql, redis, sodium, zip, Zend OPcache

**Nextcloud version:** 22.0.0 - 22.0.0.11

**Tasks version:** 0.14.0-alpha.1

**Updated from an older Nextcloud or fresh install:** updated from 21

**List of activated apps:**

Enabled:
 - accessibility: 1.7.0
 - activity: 2.15.0
 - calendar: 2.3.0
 - circles: 22.0.0
 - cloud_federation_api: 1.4.0
 - comments: 1.11.0
 - contacts: 4.0.0
 - contactsinteraction: 1.2.0
 - dashboard: 7.1.0
 - dav: 1.18.0
 - deck: 1.5.0
 - federatedfilesharing: 1.11.0
 - federation: 1.11.0
 - files: 1.16.0
 - files_pdfviewer: 2.3.0
 - files_rightclick: 1.1.0
 - files_sharing: 1.13.2
 - files_trashbin: 1.11.0
 - files_versions: 1.14.0
 - files_videoplayer: 1.11.0
[...]

PR #1694, opened 2021-07-18
Fix moving tasks back and forth

raimund-schluessler:
This is the first attempt to (hot-)fix moving tasks between lists, see #1685.

Requires https://github.com/nextcloud/cdav-library/pull/514 for actually sending the correct header to server.

Unfortunately, the server does not seem to respect neither the `X-NC-CalDAV-No-Trashbin` header nor the request to overwrite existing dav objects.

So this will definitely need fixes on the server side. Until this is fixed in server, one can only delete the trash bin entry and then move the task back.


raimund-schluessler:
I will close this since it's a server issue to be closed with https://github.com/nextcloud/server/pull/30120.

PR #1701, opened 2021-07-18
Use space after async declaration

raimund-schluessler:
Will be necessary for https://github.com/nextcloud/tasks/pull/1690


raimund-schluessler:
This needs the new eslint-rules from #1690.

PR #1704, opened 2021-07-20
l10n: Change description for alphabetical sort

Valdnet:
The description for alphabetical sort is inconsistent with the specified function.

### DIFF
--- a/src/components/SortorderDropdown.vue
+++ b/src/components/SortorderDropdown.vue
@@ -151,7 +151,7 @@ export default {
 					id: 'alphabetically',
 					icon: 'OrderAlphabeticalAscending',
 					text: this.$t('tasks', 'Alphabetically'),
-					hint: this.$t('tasks', 'Sort by summary and priority.'),
+					hint: this.$t('tasks', 'Sort by alphabet.'),
 				},
 				{
 					id: 'manual',


raimund-schluessler:
> 
> 
> The description for alphabetical sort is inconsistent with the specified function.

It's actually not. We sort alphabetically by the summary and as a second comparison (in case of equal summary) we sort by priority. See
https://github.com/nextcloud/tasks/blob/093c78e0d65601c7fcaefcd5a1b41186c31e58c4/src/store/storeHelper.js#L220-L222
https://github.com/nextcloud/tasks/blob/093c78e0d65601c7fcaefcd5a1b41186c31e58c4/src/store/storeHelper.js#L317-L319

Also compare the hint for sorting by priority:
https://github.com/nextcloud/tasks/blob/a47e7231078b9b960d73d34e35566e6d00da7609/src/components/SortorderDropdown.vue#L148

Valdnet:
@raimund-schluessler In this case, there should be no: `Sort alphabetically by summary and priority.`?
This is currently a bit confusing to the user 😉.

raimund-schluessler:
> @raimund-schluessler In this case, there should be no: `Sort alphabetically by summary and priority.`?

@Valdnet I don't really know what you mean. We sort alphabetically, and the property we sort alphabetically is the summary. And this string `Sort alphabetically by summary and priority.` is not used in the app.

Valdnet:
Hmmm. It is a bit incomprehensible. It's sort by priority and here is the description: `Sort by priority and summary.` However, for alphabetical sorting, the description is similar, only the order is changed: `Sort by summary and priority.`.
Why is the feature description: `Alphabetically`?

Summary:
For a user who does not see the application code, it is clear that sorting is done by setting tasks according to the alphabet.

PR #1775, opened 2021-09-11
Use @nextcloud/l10n everywhere

raimund-schluessler:


raimund-schluessler:
Test failure is unrelated.

Issue #1692, opened 2021-07-18
Allow to delete all items in the trash bin at once


raimund-schluessler:

Aeredren:
I coded it in frontend by iterating over the list of items, as suggested in the similar issue for calendar https://github.com/nextcloud/calendar/issues/3171

I still have to fix the button placement before any PR.

raimund-schluessler:
> 
> 
> I coded it in frontend by iterating over the list of items, as suggested in the similar issue for calendar [nextcloud/calendar#3171](https://github.com/nextcloud/calendar/issues/3171)
> 
> I still have to fix the button placement before any PR.

Great, thanks a lot! Feel free to send a (draft) PR already.

PR #1802, opened 2021-09-28
Empty trash bin

Aeredren:
fixes: https://github.com/nextcloud/tasks/issues/1692

### DIFF
--- a/src/components/AppNavigation/Trashbin.vue
+++ b/src/components/AppNavigation/Trashbin.vue
@@ -85,9 +85,14 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 								</div>
 							</template>
 						</div>
-						<p v-if="retentionDuration" class="footer">
-							{{ n('calendar', 'Elements in the trash bin are deleted after {numDays} day', 'Elements in the trash bin are deleted after {numDays} days', retentionDuration, { numDays: retentionDuration }) }}
-						</p>
+						<div class="footer">
+							<p v-if="retentionDuration">
+								{{ n('calendar', 'Elements in the trash bin are deleted after {numDays} day', 'Elements in the trash bin are deleted after {numDays} days', retentionDuration, { numDays: retentionDuration }) }}
+							</p>
+							<button @click="onEmptyTrashBin()">
+								{{ t('tasks','Empty trash bin') }}
+							</button>
+						</div>
 					</template>
 				</div>
 			</Modal>
@@ -251,6 +256,23 @@ export default {
 				showError(t('tasks', 'Could not restore calendar or event'))
 			}
 		},
+		onEmptyTrashBin() {
+			OC.dialogs.confirm(
+				t('tasks', 'Do you really want to empty the trash bin?'),
+				t('tasks', 'Empty trash bin'),
+				this.emptyTrashBin,
+				true
+			)
+		},
+
+		emptyTrashBin(confirm) {
+			if (!confirm) {
+				return
+			}
+			this.items.forEach((item) => {
+				this.onDeletePermanently(item)
+			})
+		},
 	},
 }
 </script>
@@ -315,6 +337,9 @@ export default {
 	text-align: center;
 	font-size: small;
 	margin-top: 16px;
+	& > p {
+		margin-bottom: 12px;
+	}
 }
 
 .icon-bullet {
--- a/src/components/AppNavigation/Trashbin.vue
+++ b/src/components/AppNavigation/Trashbin.vue
@@ -222,9 +222,9 @@ export default {
 					break
 				}
 			} catch (error) {
-				logger.error('could not restore ' + item.url, { error })
+				logger.error('could not delete ' + item.url, { error })
 
-				showError(t('tasks', 'Could not restore calendar or event'))
+				showError(t('tasks', 'Could not delete calendar or event'))
 			}
 		},
 		async restore(item) {


src/components/AppNavigation/Trashbin.vue
```diff
@@ -42,6 +42,14 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 					</EmptyContent>
 					<template v-else>
 						<h2>{{ t('tasks', 'Trash bin') }}</h2>
+						<ActionButton
```
raimund-schluessler, line 45: `ActionButton` needs to be wrapped inside and `Actions` component. See https://github.com/nextcloud/tasks/pull/1802/files#diff-a605b8ea178482e0a0bcab69e53d101699e820dde4ae75e617a1d993bd92c634R84-R92 for an example.

src/components/AppNavigation/Trashbin.vue
```diff
@@ -42,6 +42,14 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
...
+							@click="onEmptyTrashBin(items)">
+							<template #icon>
+								<Delete :size="20" decorative />
+							</template>
+							{{ t('tasks','Delete permanently') }}
```
raimund-schluessler, line 50:
```suggestion
							{{ t('tasks','Empty trash bin') }}
```

Aeredren:
I don't really like what I did with the button. It should have a trashbin icon imo but I could'nt find how to add it. And perhaps it's better to put it at the end :
- less misclick possible
- but necessity to scroll down the loooong trashbin history to find it

Even better would be an undo action because we're talking batch deletion here... I will look toward it this week.

raimund-schluessler:
@Aeredren Please don't commit changes to the l10n files. They are handled via transifex and generated automatically. Your changes would be overwritten on the next sync.

raimund-schluessler:
> 
> 
> I don't really like what I did with the button. It should have a trashbin icon imo but I could'nt find how to add it.
 
Yeah, that's a bit tricky at the moment, you have to do it manually. Just put the `Delete` icon component within `button` and style the icon position properly. This will be better once we have the button component from nextcloud/vue: https://github.com/nextcloud/nextcloud-vue/pull/1808


> And perhaps it's better to put it at the end :
> 
>    * less misclick possible
> 
>    * but necessity to scroll down the loooong trashbin history to find it
> 

I don't have a strong opinion here. But if you put it at the end, it should be statically visible in the footer. Otherwise users will never find it.

> Even better would be an undo action because we're talking batch deletion here... I will look toward it this week.

Well, this is the trash bin. Having an undo for the trash bin would be weird, I think, since the trash bin is the undo already, no? We could have a confirmation modal though. That's also the way every other trash bin implementation I know of does it. And I know that undo is preferred over confirmation, but here we have one of the few exceptions, I would say. What do @nextcloud/designers think here?

nimisha-vijay:
Are there any screenshots we could see?

But from what I can tell, I agree with everything, an "empty trash bin" button with a confirmation modal is good. Since it might get filled up with a lot of entries it makes sense to have it always visible. If there are 100+ entries, that's when you'll likely want to empty the bin anyway and you wouldn't want to scroll so much. A button that's fixed to the bottom of the modal sounds good.

Aeredren:
I move the button to the footer, but I could not find how to add icons nor confirmation modal.
Here a screenshot of what it look like :
trashbin_screenshot1

For the icon we can maybe wait for the Button component ?
For the modal, if anyone know how to do it feel free to implement it.

raimund-schluessler:
> For the modal, if anyone know how to do it feel free to implement it.

For the modal we will just use the Modal component from nextcloud/vue:
https://nextcloud-vue-components.netlify.app/#/Components/Modal

raimund-schluessler:
> 
> 
> > For the modal, if anyone know how to do it feel free to implement it.
> 
> For the modal we will just use the Modal component from nextcloud/vue: https://nextcloud-vue-components.netlify.app/#/Components/Modal

Although the trash bin is a modal itself already, let's see how that works out 🙈

Aeredren:
> Although the trash bin is a modal itself already, let's see how that works out see_no_evil

That's why I did'nt want to try and implement it, look like problems are coming ://
I might give it a try this afternoon if I have some time.

raimund-schluessler:
> 
> 
> > Although the trash bin is a modal itself already, let's see how that works out see_no_evil
> 
> That's why I did'nt want to try and implement it, look like problems are coming :// I might give it a try this afternoon if I have some time.

It seems to be fine. Checkout this code:
```vue
<template>
	<div>
		<button @click="showModal1">Show Modal1</button>
		<modal v-if="modal1" @close="closeModal1" size="small">
			<div class="modal__content">Trash bin</div>
			<button @click="showModal2">Empty</button>
		</modal>
		<modal v-if="modal2" @close="closeModal2" size="small">
			<div class="modal__content">Your really want to do this?</div>
			<button>Yes</button>
			<button>No</button>
		</modal>
	</div>
</template>
<style scoped>
	.modal__content {
		margin: 50px;
		text-align: center;
	}

</style>
<script>
export default {
	data() {
		return {
			modal1: false,
			modal2: false,
		}
	},
	methods: {
		showModal1() {
			this.modal1 = true
		},
		closeModal1() {
			this.modal1 = false
		},
		showModal2() {
			this.modal2 = true
		},
		closeModal2() {
			this.modal2 = false
		}
	}
}
</script>
```
Copy it into the code input field on this page https://nextcloud-vue-components.netlify.app/#/Components/Modal to try it out.

Aeredren:
I was feeling like reinventing the wheel with this confirmation modal. And as [no vue composant for confirmation exist yet](https://github.com/nextcloud/nextcloud-vue/issues/408) in nextcloud composant, I try to reuse the legacy OC.dialogs.confirm. Here what it look like :
trashbin_screenshot2

Is it a good idea to use this legacy code or not ? If not we could wait for the vue component or else I have a draft for the modal in enh/emptyTrash_modal. (But I'm not good at css ^^)

src/components/AppNavigation/Trashbin.vue
```diff
@@ -251,6 +267,15 @@ export default {
 				showError(t('tasks', 'Could not restore calendar or event'))
 			}
 		},
+		async onEmptyTrashBin(items) {
```
raimund-schluessler, line 270:
```suggestion
		onEmptyTrashBin() {
			OC.dialogs.confirm(
				t('tasks', 'Do you really want to empty the trash bin?'),
				t('tasks', 'Empty trash bin'),
				this.emptyTrashBin,
				true
			)
		},
		emptyTrashBin(confirm) {
			if (!confirm) {
				return
			}
			this.items.forEach((item) => {
				this.onDeletePermanently(item)
			})
		},
```

src/components/AppNavigation/Trashbin.vue
```diff
@@ -85,9 +85,14 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
...
 							</template>
 						</div>
-						<p v-if="retentionDuration" class="footer">
-							{{ n('calendar', 'Elements in the trash bin are deleted after {numDays} day', 'Elements in the trash bin are deleted after {numDays} days', retentionDuration, { numDays: retentionDuration }) }}
-						</p>
+						<div class="footer">
+							<p v-if="retentionDuration">
+								{{ n('calendar', 'Elements in the trash bin are deleted after {numDays} day', 'Elements in the trash bin are deleted after {numDays} days', retentionDuration, { numDays: retentionDuration }) }}
+							</p>
+							<button @click="onEmptyTrashBin(items)">
```
raimund-schluessler, line 92:
```suggestion
							<button @click="onEmptyTrashBin()">
```

raimund-schluessler:
> Is it a good idea to use this legacy code or not ? If not we could wait for the vue component or else I have a draft for the modal in [enh/emptyTrash_modal](https://github.com/Aeredren/tasks/tree/enh/emptyTrash_modal).

I don't really know. Other apps also still use OC.dialogs, so I guess it is fine.

I suggest to simplify the code a bit and create a second method instead of making the dialog a promise. I couldn't push to your branch, so please checkout the code suggestions above.

Besides that, it works nicely.

src/components/AppNavigation/Trashbin.vue
Aeredren: Done.

src/components/AppNavigation/Trashbin.vue
Aeredren: Done

Aeredren:
I guess I'll port this to the Calendar trash bin once this PR is good enough to be merge.

raimund-schluessler:
Works well now and the code looks good. 👍 

We should probably only run 5 delete requests at a time to not overload the server when many elements are to be deleted. But this can be done in a follow-up PR.

Issue #89, opened 2017-06-24
Improvements for categories


spoorun:
Nextcloud has a tagging system.
'Categories' are tags, and should sync with that system.

The benefits would be: 
1) Users managing different types of data (such as tasks, contacts, messages, files) would use the same sets of tags (including shared tags) to organise that data, 
2) and in future be able to find that data by global searches. 
3) It would make shared data management more consistent too.
4) The UI for tags would be consistent across apps (and that could be improved), and it would be more efficient - Users shouldn't need to learn endless different UIs and modes of operation to carry out the same task (using tags) within a single system (Nextcloud).

raimund-schluessler:
I will use this issue to collect all the improvements requested for categories:

- [ ] Integrate into Nextcloud tagging system
- [ ] Allow adding categories / tags by hashtagging in the title (https://github.com/nextcloud/tasks/issues/42)
- [ ] Integrate into search, clicking a tag should start a search (https://github.com/nextcloud/tasks/issues/47)

websafe:
@raimund-schluessler  any update on this enhancement?

udolee:
Tasks is so good! One question, is there a way to sort by categories? Thanks!

GambaJo:
Yap, this is very pitty. I miss this feature too.

klawdhfzasjhaa:
> Tasks is so good! One question, is there a way to sort by categories? Thanks!

Additionally, it would be great to filter by category.

javisomoza:
Yeah, i miss also the feature of filtering by category. And the category tags being also links to that filtering.

duesseldorferjung:
+1

GambaJo:
> 
> 
> > Tasks is so good! One question, is there a way to sort by categories? Thanks!
> 
> Additionally, it would be great to filter by category.

+1

abers:
> Yeah, i miss also the feature of filtering by category. And the category tags being also links to that filtering.

I came to the git repo to see if anyone had requested this feature already. Do you know why it was removed? It'd be great if could view list of categories in left-hand panel, filter by category, and view all tasks tagged with a category. Would replicate the main features from Tasks.org (mobile app I sync tasks with) that are currently missing from Tasks.

raimund-schluessler:
> > Yeah, i miss also the feature of filtering by category. And the category tags being also links to that filtering.
> 
> I came to the git repo to see if anyone had requested this feature already. Do you know why it was removed?

[...]

PR #1836, opened 2021-11-03
[WIP] Slightly automated tagging for new Tasks

caemor:
Currently tags, calendar and date can't be set from the input field directly AFAIK.
This PR is work in progress. Currently only the auto-tagging should work but even that features hasn't been tested yet.

 Current ideas and questions for this PR:
 * Tags should be annotated by an '@'. 
 * Calendar should be set with an '#'. How to get a correct calendar? 
 * Due Date is even more WIP. Since every special case needs to be handled
   Maybe simplify it to: today, tomorrow, next week (Monday), next month or dd/mm or dd/mm/yyyy for now
 * Auto-matching for tags and calendar? Is auto-matching possible? How to get the correct calendar or a tag (e.g. "work" vs "Work",...)
   
 Current status:
- [X] Tagging (not yet tested!)
- [ ] Test Tagging
- [ ] Calendar
- [ ] Due Date
- [ ] ...

This is related to #89

### DIFF
--- a/src/store/tasks.js
+++ b/src/store/tasks.js
@@ -308,7 +308,7 @@ const mutations = {
 	 * @param {Array<Task>} tasks Tasks
 	 */
 	appendTasks(state, tasks = []) {
-		state.tasks = tasks.reduce(function(list, task) {
+		state.tasks = tasks.reduce(function (list, task) {
 			if (task instanceof Task) {
 				Vue.set(list, task.key, task)
 			} else {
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
@caemor Thanks for working on this.

I would prefer to handle the different features in separate PRs and focus on the tagging feature first (since it is the easiest).

- Please use the hashtag `#` as the annotation for tags. This is the most common character for this (see Twitter) and likely what the user expects.
- Parsing the user input should go into a separate function for easier testing. The function should accept a string and return an object with the properties `title<string>` and `tags<array>`.

As for the other features (in separate PRs):

Calendar:
- The `@` character makes sense for indicating where a task gets created (i.e. @/at a certain place/list). So please use it for indicating in which calendar/list a task will reside.
- Overwriting the calendar needs a bit of thinking. We normally show in the input field in which Calendar a task gets created. Overwriting this might lead to unexpected results. And we should for sure forbid to create a subtask in a different calendar, as it will not be visible after creation. Same goes for being in a list view.

Dates:
- There are really many key-words to parse for. I would start with today, tomorrow, ISO-dates (e.g. 2021-11-05) and the date format of the user. Next + weekday would also be an option, but this gets tricky already.

jayarmstrong:
Thank you for starting this!

@ are normally used to tag people / groups. In the interest of that also being used in the future, I'd suggest using something else for adding calendars. Unless calendars, people, and groups are equivalent classes.

Toodledo has implemented this extremely well, IMO. They use a `/` in the task title to add details. This is the fastest way to add tasks that I've come across and doesn't require the constant keyboard-to-mouse switching like NC and most other task managers. I won't detail all its greatness here but here are some basic examples:
- `task 1 /mon` = set due date OR start date to following  Monday
- `task 1 /high /coding /paid` = set priority high, add tags coding and paid
- `task 1 /groceries` = add tag "groceries"
Of course you lose or slow down the ability to use these tag delimiters (/, #, @, etc) so that's one reason to bundle some / all of the tagging into fewer delimiters. Personally, I'd like to see @ for involving people and groups, / for details and tags, and optionally # for tags only.

raimund-schluessler:
Superseded by #2425.

PR #1916, opened 2022-01-27
Fix: Inconsistencies due to ALTREP parameter

max65482:
This PR addresses an inconsistency issue with Thunderbird. In addition to the plain text description of a task, Thunderbird also saves a formatted HTML version inside the ALTREP parameter. NC calendar does not alter the ALTREP parameter when the plain text description is changed. This results in inconsistencies.

The proposed solution deletes the ALTREP parameter upon modification. This prevents inconsistencies. Thunderbird keeps accepting plaintext-only descriptions.

I have opened a [similar PR](https://github.com/nextcloud/calendar/pull/3918) in nextcloud/calendar as it is equally concerned.

### DIFF
--- a/src/models/task.js
+++ b/src/models/task.js
@@ -366,7 +366,7 @@ export default class Task {
 	}
 
 	set note(note) {
-		// To avoid inconsistent parameters, delete complete property, then recreate
+		// To avoid inconsistent property parameters (bug #3863 in nextcloud/calendar), delete the property, then recreate
 		this.vtodo.removeProperty('description')
 		this.vtodo.addPropertyWithValue('description', note)
 		this.updateLastModified()
--- a/src/models/task.js
+++ b/src/models/task.js
@@ -367,6 +367,7 @@ export default class Task {
 
 	set note(note) {
 		this.vtodo.updatePropertyWithValue('description', note)
+		this.vtodo.getFirstProperty('description').removeParameter('altrep')
 		this.updateLastModified()
 		this._note = this.vtodo.getFirstPropertyValue('description') || ''
 	}
--- a/src/models/task.js
+++ b/src/models/task.js
@@ -366,8 +366,9 @@ export default class Task {
 	}
 
 	set note(note) {
-		this.vtodo.updatePropertyWithValue('description', note)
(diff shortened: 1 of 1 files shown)


benbucksch:
Thanks, max, for the patch!

As mentioned in https://github.com/nextcloud/calendar/issues/3863#issuecomment-1007533913 , I would recommend to delete all unknown parameters. They give additional information about the description, but the description changed, so any parameter that is still here is in risk of being out of sync now.

src/models/task.js
```diff
@@ -366,7 +366,9 @@ export default class Task {
 	}
 
 	set note(note) {
-		this.vtodo.updatePropertyWithValue('description', note)
+		// To avoid inconsistent parameters, delete complete property, then recreate
```
benbucksch, line 369:
```suggestion
		// To avoid inconsistent property parameters (bug #3863), delete the property, then recreate
```

benbucksch (commented):
I like the idea

src/models/task.js
max65482: Done! (Add: issue is in nextcloud/calendar)

Gorendal:
Hi there!
Looks like the X-ALT-DESC parameter is not yet cleared for Tasks like for Calendar items. (see https://github.com/nextcloud/calendar/issues/3863#issuecomment-1315672382). 
I guess you would like me to create a pull-request again - but I still don't know how.
Could you please support?

max65482:
Do you want to report the problem? Then you can file an issue (click "New issue" [here](https://github.com/nextcloud/tasks/issues) and fill in the gaps).
If you want to implement the solution yourself maybe look at [#4744](https://github.com/nextcloud/calendar/pull/4744) (the fix for calendar) and #1916 (a similar fix in tasks).

Gorendal:
Hi Max!
Sorry for the late reply. I hoped like the similar fix for Nextcloud Calendar this might not require all the efforts for creating a new issue. However as you see above I created one right now.
Unfortunatelly I have no Nextcloud development environment set up and only little Github knowledge so far. So I am not able to fix this by myself without excessive preparation efforts. So I hope you or somebody else might help here?

max65482:
Fix is in #2240.

PR #1971, opened 2022-04-22
chore: migrate the app to vue 3

raimund-schluessler:
This PR migrates the Tasks app to use vue3.

I would like to merge this issue even though we do not have a final release of `nextcloud/vue@9` yet. It has been open for long enough, and we should push towards a final release anyway. In case we do not find any issues, I think it should be ok to merge already.


package.json
```diff
@@ -31,14 +31,14 @@
     "@nextcloud/axios": "^2.4.0",
     "@nextcloud/calendar-js": "6.1.0",
     "@nextcloud/cdav-library": "1.1.0",
-    "@nextcloud/dialogs": "^5.1.1",
+    "@nextcloud/dialogs": "4.1.0",
```
susnux, line 34: Have you found issues using v5 with a vue app? Would be interested :)

raimund-schluessler: No, there are no issues. The downgrade is just necessary at the moment because since version `4.2.0` `nextcloud/dialogs` depends on `nextcloud/vue@8` and `vue2`, so it needs to be migrated to vue 3 first. But that is not an issues here, because the Tasks app only uses `showError` from the lib which works the same in `4.1.0`.

package.json
```diff
@@ -51,22 +51,20 @@
...
     "uuid": "^9.0.1",
-    "vue": "^2.7.16",
+    "vue": "^3.4.15",
     "vue-material-design-icons": "^5.2.0",
-    "vue-router": "^3.6.5",
-    "vuedraggable": "^2.24.3",
-    "vuex": "^3.6.2",
-    "vuex-router-sync": "^5.0.0"
+    "vue-router": "^4.2.5",
+    "vuex": "^4.1.0"
```
skjnldsv, line 59: pinia?

skjnldsv (approved):

skjnldsv:
Time for 1.0.0? :grin: 

image

package.json
raimund-schluessler: Yes, but likely as a follow-up.

raimund-schluessler:
> Time for 1.0.0? 😁
> 
> image

🙂
My internal planning was to release 1.0.0 once we have #2318.

susnux (approved):
really nice work!

PR #1988, opened 2022-05-06
add note for tasks app

portno12:
this app requires a subscription to sync with nextcloud and should be labelled for clarity.
not sure how to DCO via github web, feel free to close this and recreate if needed

docs:https://tasks.org/docs/sync


raimund-schluessler:
For such a small change, I think we can go without DCO. Thanks.

PR #1998, opened 2022-05-09
Display tasks with future start date dimly

nico-nacq:
Render future tasks (tasks with a start date in the future) differently than other tasks : Lower opacity, little icon on the right. Italic text.

Before : 

After : 

What changes :

EDIT : After a few tweeks suggested by @raimund-schluessler, here's the result : 

(no italic, better icon and alt text)

(This is the first time I make a PR. Hope I'm doing it right 🤞)

Thank you 😊

### DIFF
--- a/src/components/TaskBody.vue
+++ b/src/components/TaskBody.vue
@@ -656,16 +656,18 @@ $breakpoint-mobile: 1024px;
 		opacity: .6;
 	}
 
-	&--non-started .title {
-		color: var(--color-placeholder-dark);
-	}
+	&--non-started {
+		.title {
+			color: var(--color-placeholder-dark);
+		}
 
-	&--non-started > .task-item__body {
-		background-color: var(--color-background-dark);
-	}
+		> .task-item__body {
+			background-color: var(--color-background-dark);
+		}
 
-	&--non-started .task-checkbox {
-		opacity: .2;
+		.task-checkbox {
+			opacity: .2;
+		}
 	}
 
 	&.sortable-ghost {
--- a/src/components/TaskBody.vue
+++ b/src/components/TaskBody.vue
@@ -76,6 +76,7 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 					<span class="calendar__name">{{ task.calendar.displayName }}</span>
 				</div>
 				<SortVariant v-if="hasHiddenSubtasks" :size="20" :title="t('tasks', 'Task has hidden subtasks')" />
+				<CalendarClock v-if="!overdue(task.startMoment) && task.start" :size="20" :title="t('tasks', 'Task has not yet started')" />
 				<Pin v-if="task.pinned" :size="20" :title="t('tasks', 'Task is pinned')" />
 				<TextBoxOutline v-if="task.note!=''"
 					:size="20"
@@ -182,6 +183,7 @@ import Pin from 'vue-material-design-icons/Pin'
 import Plus from 'vue-material-design-icons/Plus'
 import TextBoxOutline from 'vue-material-design-icons/TextBoxOutline'
 import SortVariant from 'vue-material-design-icons/SortVariant'
+import CalendarClock from 'vue-material-design-icons/CalendarClock'
 import Star from 'vue-material-design-icons/Star'
 import Undo from 'vue-material-design-icons/Undo'
 
@@ -206,6 +208,7 @@ export default {
 		Plus,
 		TextBoxOutline,
 		SortVariant,
+		CalendarClock,
 		Star,
 		Undo,
 	},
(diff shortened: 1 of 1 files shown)


src/components/TaskBody.vue
```diff
@@ -652,6 +653,27 @@ $breakpoint-mobile: 1024px;
 		opacity: .6;
 	}
 
+	&--non-started .title {
+		color: var(--color-placeholder-dark);
+		font-style: italic;
+	}
+
+	&--non-started .title:before {
+		content: '⏱️';
```
raimund-schluessler, line 662: Please use material design icon for this. Just put it next to the `Notes` and `Pin` icons here:
https://github.com/nextcloud/tasks/blob/00dc26d40aec882196a3cf56bd065d6000532e1f/src/components/TaskBody.vue#L77-L82

raimund-schluessler: The clock icon would be suitable, I think:
https://materialdesignicons.com/icon/alarm

src/components/TaskBody.vue
```diff
@@ -652,6 +653,27 @@ $breakpoint-mobile: 1024px;
 		opacity: .6;
 	}
 
+	&--non-started .title {
+		color: var(--color-placeholder-dark);
+		font-style: italic;
```
raimund-schluessler, line 658: Not so sure about the italic, we don't use it very often in Nextcloud. I would tend to remove it.

raimund-schluessler:
@nico-nacq Thanks for your PR. I left a few comments. In principle I am fine with changing the appearance of future tasks, but I would like to hear @nextcloud/designers about this as well.

src/components/TaskBody.vue
nico-nacq: > Please use a material design icon for this. Just put it next to the `Notes` and `Pin` icons

Yes of course : 5243df0 ... Thanks for pointing me the right direction !

> The clock icon would be suitable, I think

If I may, I prefere the `calendar-clock`
image

I think it looks more consistent with the `calendar-start` used in the task settings :
image

It looks like this (with the alt text) :

src/components/TaskBody.vue
nico-nacq: Done !

a191d5e

nico-nacq:
> @nico-nacq Thanks for your PR. I left a few comments. In principle I am fine with changing the appearance of future tasks, but I would like to hear @nextcloud/designers about this as well.

Thank you so much for considering my PR and for your work !

src/components/TaskBody.vue
raimund-schluessler: > > The clock icon would be suitable, I think
> 
> If I may, I prefere the `calendar-clock` image

Fine with me.

src/components/TaskBody.vue
```diff
@@ -652,6 +656,18 @@ $breakpoint-mobile: 1024px;
...
+	}
+
+	&--non-started > .task-item__body {
+		background-color: var(--color-background-dark);
+	}
+
+	&--non-started .task-checkbox {
+		opacity: .2;
+	}
+
```
raimund-schluessler, line 670:
```suggestion
	&--non-started {
		.title {
			color: var(--color-placeholder-dark);
		}
		> .task-item__body {
			background-color: var(--color-background-dark);
		}
		.task-checkbox {
			opacity: .2;
		}
	}
```

jancborchardt:
Nice, good idea!

Just some concern regarding accessibility and readability: We do have the variable `--color-text-maxcontrast` which is subdued but ensures proper contrast, but it has to be on `--color-main-background` (not grey as well) to be readable. Would be better to use that, and the text + checkbox color adjustment is probably enough?

raimund-schluessler:
> Nice, good idea!
> 
> Just some concern regarding accessibility and readability: We do have the variable `--color-text-maxcontrast` which is subdued but ensures proper contrast, but it has to be on `--color-main-background` (not grey as well) to be readable. Would be better to use that, and the text + checkbox color adjustment is probably enough?

Would look like this (second task starts in the future, icon omitted, since it's only done in devtools):
Screenshot 2022-05-16 at 14-49-35 Aufgaben - Nextcloud

raimund-schluessler:
I would prefer it without a background as well. If one wants a different background color, it will be very simple to change it with the custom CSS app, since we now set a CSS class for tasks starting in the future with this PR.

nico-nacq:
> Just some concern regarding accessibility and readability: We do have the variable `--color-text-maxcontrast` which is subdued but ensures proper contrast, but it has to be on `--color-main-background` (not grey as well) to be readable. Would be better to use that, and the text + checkbox color adjustment is probably enough?

Ok done : a1abc27

 - The background stays unchanged.
 - The text goes to `--color-text-maxcontrast`.
 - The checkbox still fades to .2
 - The icon is still added on the right.

It looks like this  :

nico-nacq:
Thank you 🙏🥲

jancborchardt:
Looks great, nice work @nico-nacq! Looking forward to more design contributions from you if you like. :) Check https://nextcloud.com/design for more info

PR #2016, opened 2022-05-29
Move SCSS into components

raimund-schluessler:
Since Nextcloud server 25 will stop compiling SCSS files, we have to compile it ourself. Hence, we move the SCSS code into the vue components.

@skjnldsv What is left is including icons for circles in
https://github.com/nextcloud/tasks/blob/f4738a154c9bb2412bb4ea2ede38922a372cfd43/css/tasks.scss#L1

Any recommendation how to do this for Nextcloud 25?


raimund-schluessler:
@skjnldsv I just saw, server provides an `icon-circles` class with the correct icon. So I can use this.
But for
https://github.com/nextcloud/tasks/blob/f4738a154c9bb2412bb4ea2ede38922a372cfd43/css/tasks-talk.scss#L1
the question remains, I guess.

raimund-schluessler:
Merging, since otherwise Tasks is not usable with NC25. More fixes/adjustments in a follow-up, if necessary.

skjnldsv:
> @skjnldsv I just saw, server provides an `icon-circles` class with the correct icon. So I can use this. But for
> 
> https://github.com/nextcloud/tasks/blob/f4738a154c9bb2412bb4ea2ede38922a372cfd43/css/tasks-talk.scss#L1
> 
> 
> the question remains, I guess.

I'd say move it to server.
Where is this being used?

skjnldsv:
> In the Talk integration menu. But I guess that anyway needs adjustments for md icons.

Yes, or a wider discussion about how we do icons in APIs :)

PR #2073, opened 2022-08-15
Adjust server versions to be tested

raimund-schluessler:
This adjusts the nextcloud server and php versions to be tested:
- stable22 and older: EOL -> not tested
- stable23: php 7.3/7.4/8.0 (https://docs.nextcloud.com/server/23/admin_manual/installation/system_requirements.html#server)
- stable24: php 7.4/8.0/8.1 (https://docs.nextcloud.com/server/24/admin_manual/installation/system_requirements.html#server)
- master: same as stable24


raimund-schluessler:
@ChristophWurst I am a bit stuck here, since I can't get the composer.json and composer.lock.json to work with PHP 7.3 to 8.1. It always complains about `Your lock file does not contain a compatible set of packages. Please run composer update.`, but that fixes it only for the PHP version I run composer update for. Any hint would be very appreciated.

ChristophWurst:
Looks like your symfony/* packages are too new. Try overwriting the php version in composer.json's config like https://github.com/nextcloud/calendar/blob/main/composer.json#L4, then run `composer update`. This will hopefully give you a set of packages that work from 7.4 to 8.1. IIRC we had rare cases where this was impossible. Then we installed packages dynamically on CI or used the `--ignore-platform-req=php` flag of `composer i`.

raimund-schluessler:
> Looks like your symfony/* packages are too new. Try overwriting the php version in composer.json's config like https://github.com/nextcloud/calendar/blob/main/composer.json#L4, then run `composer update`. This will hopefully give you a set of packages that work from 7.4 to 8.1. IIRC we had rare cases where this was impossible. Then we installed packages dynamically on CI or used the `--ignore-platform-req=php` flag of `composer i`.

Thanks a lot, that was it.

PR #2112, opened 2022-09-17
double click to edit title and notes more easily

blairun:
1- double click task body to place cursor in task edit pane for quicker editing. 
2- double click notes icon in task body to place cursor in notes field for quicker editing.

### DIFF
--- a/src/components/TaskBody.vue
+++ b/src/components/TaskBody.vue
@@ -80,7 +80,8 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 				<TextBoxOutline v-if="task.note!=''"
 					:size="20"
 					:title="t('tasks', 'Task has a note')"
-					@click.stop="openAppSidebarTab($event, 'app-sidebar-tab-notes')" />
+					@click.stop="openAppSidebarTab($event, 'app-sidebar-tab-notes')"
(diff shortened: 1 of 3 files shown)


src/components/TaskBody.vue
```diff
@@ -578,6 +582,23 @@ export default {
 			this.navigate($event, tab)
 			// In case it is already open, we also have to emit an event to show the tab
 			emit('tasks:open-appsidebar-tab', { tab })
+			this.clicks++
```
raimund-schluessler, line 585: I think it's fine to simply bind a `@dlbclick` handler here
https://github.com/nextcloud/tasks/blob/868cd6d072c3f68ee4f8008bd06f381457ca5430/src/components/TaskBody.vue#L83
as well, slightly change the signature of the function called to `openAppSidebarTab($event, tab, edit)` and `emit('tasks:edit-appsidebar-notes', $event)` if `edit` is true. That saves us the custom double click implementation. If you double click then, the sidebar tab is technically opened three times, but that doesn't hurt (it's opened twice with the current code already).

raimund-schluessler: @blairun I added a commit to your branch. Please check whether that works for you. Then we can merge the PR.

blairun: Those changes work well for me.  Thanks!

PR #2117, opened 2022-09-23
Use global import for nextcloud-vue

raimund-schluessler:
Signed-off-by: Raimund Schlüßler <raimund.schluessler@mailbox.org>


raimund-schluessler:
The build size still slightly increases when using the global import. Closing.

PR #2209, opened 2023-02-28
Show notes equally in editing and preview mode

raimund-schluessler:


raimund-schluessler:
> I just got the update in my NC instance and it looks mostly great. Linefeeds are still inconsistent. For example "\n\n" in edit mode is trimmed down to "\n".

Yes, as far as I can tell, that is a limitation of markdown itself. See the statement of the creator of the mardown plugin we use:
https://github.com/markdown-it/markdown-it/issues/211#issuecomment-191932958

Issue #1894, opened 2021-12-22
[Feature Request] Add multiple tasks at once (bulk / batch)


CodeAlDente:
I'd like to add multiple tasks in one step like in Google Tasks. So if I have multiple lines (one line = one task) in my clipboard and just paste it in the field it should offer me to add them all in a row.
Maybe even with the detection of intended lines to automatically handle sub tasks?

Example with multiple tasks
```
Buy milk
Walk the dog
Clean living room
```

Example with sub tasks
```
Go to supermarket
  buy milk
  buy cheese
Invite friends to party
Call Bob
```

Redsandro:
I just tried out Tasks and this was my first annoyance with the current implementation. I'm surprised you don't have more responses or thumbs up.

At least you can keep typing when pressing enter, that helps. But it would be great if you can copy/paste lists from text files or emails. That includes removing the bullet point, e.g.: `-` or `*`.

PR #2273, opened 2023-05-21
Allow bulk task creation on paste

t-h-e:
Solves issue https://github.com/nextcloud/tasks/issues/1894

See tests in tests/javascript/unit/utils/textToTask.spec.js for all possible list structures that can be pasted to create multiple tasks in one go

### DIFF
--- a/src/components/CreateMultipleTasksDialog.vue
+++ b/src/components/CreateMultipleTasksDialog.vue
@@ -27,8 +27,7 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 				<NcButton @click="cancel">
 					{{ t('tasks', 'Cancel') }}
 				</NcButton>
-				<NcButton
-					type="primary"
+				<NcButton type="primary"
 					@click="addTasks">
 					{{ t('tasks', 'Create tasks') }}
 				</NcButton>
@@ -63,8 +62,6 @@ import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import NcModal from '@nextcloud/vue/dist/Components/NcModal.js'
 
 import Check from 'vue-material-design-icons/Check.vue'
-import TextBoxOutline from 'vue-material-design-icons/TextBoxOutline.vue'
-import ViewHeadline from 'vue-material-design-icons/ViewHeadline.vue'
 
 import { mapActions } from 'vuex'
 
@@ -76,26 +73,26 @@ export default {
 		NcEmptyContent,
 		NcLoadingIcon,
 		NcModal,
-		TextBoxOutline,
-		ViewHeadline,
 	},
 	props: {
 		calendar: {
 			type: Object,
-			required: true
+			required: true,
 		},
 		tasksToCreate: {
 			type: Object,
-			required: true
+			required: true,
 		},
 		tasksAdditionalProperties: {
 			type: Object,
-			default: {}
+			default() {
+				return {}
+			},
 		},
 		rootTask: {
 			type: Object,
-			default: undefined
-		}
+			default: undefined,
+		},
 	},
 	emits: ['cancel', 'close'],
 	data() {
@@ -130,7 +127,7 @@ export default {
 
 		async addTasks() {
 			this.creating = true
-			Promise.all(this.tasksToCreate.tasks.map(task => this.addTaskWithParent(task, this.rootTask?.uid)))
+			await Promise.all(this.tasksToCreate.tasks.map(task => this.addTaskWithParent(task, this.rootTask?.uid)))
 				.then(() => {
 					this.creating = false
 					this.created = true
@@ -142,7 +139,7 @@ export default {
 				summary: task.title,
 				calendar: this.calendar,
 				related: parentUid,
-				...this.tasksAdditionalProperties
+				...this.tasksAdditionalProperties,
 			})
 			await Promise.all(task.children.map(child => this.addTaskWithParent(child, newParent?.uid)))
 		},
@@ -190,23 +187,6 @@ export default {
 			position: absolute;
 			top: 14px;
 			left: 14px;
-		}
-
-		input,
-		textarea {
-			width: 100%;
-			padding-left: 44px !important;
-		}
-
-		input {
-			height: 44px !important;
-			margin: 0;
-		}
-
-		textarea {
-			min-width: 100%;
-			max-width: 100%;
-			min-height: 100px;
 		}
 	}
 
--- a/src/components/CreateMultipleTasksDialog.vue
+++ b/src/components/CreateMultipleTasksDialog.vue
@@ -23,18 +23,12 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 
 			<p>{{ t('tasks', 'Create {numberOfTasks} tasks from pasted text', { numberOfTasks: tasksToCreate.numberOfTasks }) }}</p>
 
-			<!-- Need to prevent keydown and keyup of first button.
-				The issue is that the first button will have focus and enter will trigger the button with keydown.
-				The input fields in HeaderBar and TaskBody which will get focus after closing the modal react on keyup.
-				Therefore an empty task would be created.
-			-->
 			<div class="modal-buttons">
-				<NcButton @click="cancel"
-					@keydown.prevent
-					@keyup.prevent>
+				<NcButton @click="cancel">
 					{{ t('tasks', 'Cancel') }}
 				</NcButton>
-				<NcButton type="primary"
+				<NcButton ref="createButton"
+					type="primary"
 					@click="addTasks">
 					{{ t('tasks', 'Create tasks') }}
 				</NcButton>
@@ -113,6 +107,10 @@ export default {
 		createdMessage() {
 			return t('tasks', '{numberOfTasks} tasks have been added to "{calendar}"', { numberOfTasks: this.tasksToCreate.numberOfTasks, calendar: this.calendar.displayName }, undefined, { sanitize: false, escape: false })
 		},
+	},
+
+	mounted() {
+		this.$nextTick(() => this.$refs.createButton?.$el?.focus())
 	},
 
 	methods: {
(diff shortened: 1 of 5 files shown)


t-h-e:
Here is an example:
BatchTaskCreation

raimund-schluessler:
@t-h-e Thank you very much. That looks very nice. The code looks good so far and I love the fact that you even provide a test. There are some lint errors that you can mostly fix with `npm run lint:fix` and I have some nitpicks. I still need to test it live though.

src/components/CreateMultipleTasksDialog.vue
```diff
@@ -0,0 +1,222 @@
...
+		},
+
+		close() {
+			this.$emit('close')
+			this.$root.$emit('close')
+		},
+
+		async addTasks() {
+			this.creating = true
+			Promise.all(this.tasksToCreate.tasks.map(task => this.addTaskWithParent(task, this.rootTask?.uid)))
```
raimund-schluessler, line 133: Could you please also use `await` here, like a few lines below?

src/components/HeaderBar.vue
```diff
@@ -91,30 +106,67 @@ export default {
...
+		addTaskWithName(task) {
 			// If the task is created in the calendar view,
 			// we set the current calendar
 			if (this.$route.params.calendarId) {
 				task.calendar = this.calendar
 			}
 
+			return this.createTask({
+				...task,
+				...this.getAdditionalTaskProperties()
```
raimund-schluessler, line 118: This is a computed prop, I don't think you need to call it with `()`, no?

src/components/CreateMultipleTasksDialog.vue
t-h-e: added

src/components/HeaderBar.vue
t-h-e: This is actually a method, see https://github.com/nextcloud/tasks/pull/2273/files#diff-6648f3fa5be4b95d4036efb24c229f414ab0f55274f6ecc7c150708d90d674b5R121

t-h-e:
Thanks for the comments. I configured eslint to be run by my IDE

src/components/TaskBody.vue
```diff
@@ -156,7 +156,8 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 					:trailing-button-label="subtasksCreationPlaceholder"
 					@trailing-button-click="addTask"
 					@keyup.esc="showSubtaskInput = false"
-					@keyup.enter="addTask">
+					@keyup.enter="addTask"
+					@paste="addMultipleTasks">
```
raimund-schluessler, line 160: Then we don't need to stop the event in the function.
```suggestion
					@paste.stop="addMultipleTasks">
```

src/components/TaskBody.vue
```diff
@@ -604,26 +619,61 @@ export default {
...
 
+		addMultipleTasks($event) {
+			const pastedText = $event.clipboardData.getData('text')
+			const tasksFromText = textToTask(pastedText)
+
+			if (tasksFromText.numberOfTasks <= 1) {
+				return
+			}
+
+			$event.stopPropagation()
```
raimund-schluessler, line 630:
```suggestion
```

src/components/HeaderBar.vue
```diff
@@ -91,30 +105,67 @@ export default {
...
+		addMultipleTasks($event) {
+			const pastedText = $event.clipboardData.getData('text')
+			const tasksFromText = textToTask(pastedText)
+
+			if (tasksFromText.numberOfTasks <= 1) {
+				return
 			}
 
-			this.createTask(task)
+			$event.stopPropagation()
```
raimund-schluessler, line 147:
```suggestion
```

src/components/HeaderBar.vue
```diff
@@ -32,11 +32,18 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 				:trailing-button-label="placeholder"
 				@trailing-button-click="addTask"
 				@keyup.esc="clearNewTask($event)"
-				@keyup.enter="addTask">
+				@keyup.enter="addTask"
+				@paste="addMultipleTasks">
```
raimund-schluessler, line 36:
```suggestion
				@paste.stop="addMultipleTasks">
```

src/components/CreateMultipleTasksDialog.vue
```diff
@@ -0,0 +1,202 @@
...
+			await Promise.all(this.tasksToCreate.tasks.map(task => this.addTaskWithParent(task, this.rootTask?.uid)))
+				.then(() => {
+					this.creating = false
+					this.created = true
+				})
+		},
+
+		async addTaskWithParent(task, parentUid) {
+			const newParent = await this.createTask({
+				summary: task.title,
```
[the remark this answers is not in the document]
raimund-schluessler: Can we please rename this property to `summary`, that's how it's called in the CalDAV spec. We try to use `title` only when related to the native HTML title attribute.

src/utils/textToTask.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ */
+
+export const textToTask = (text) => {
+	const lines = text.split(/\r?\n/)
+		.filter(isNotOnlyWhitespace)
+		.map(spaceToTab)
+
+	const rootTask = {
+		title: 'ROOT',
```
raimund-schluessler, line 27:
```suggestion
		summary: 'ROOT',
```
and all occurences below as well, please.

raimund-schluessler:
@t-h-e Thanks a lot! I have a few minor code style related comments, and will try to find the time to test the functionality later this week.

t-h-e:
I have addressed your review comments. Let me know if there is anything else 😃

raimund-schluessler:
Sorry that it takes me so long. My Nextcloud dev env was broken, and it took me a while to fix it. I will hopefully have a look at the weekend latest.

t-h-e:
Hi @raimund-schluessler just wanted to check in what the status is.
Side note: I run my Nextcloud dev setup in docker. So if there is an issue with Nextcloud, I can just remove the container and create a new one.

raimund-schluessler:
@jancborchardt Could you please add @t-h-e to nextcloud/tasks?

raimund-schluessler:
@t-h-e Thanks a lot for the contribution! I finally had time to the PR and it seems to work really nicely. The only problem I found is this:
- Paste multiple tasks into the input field
- See the `CreateMultipleTasksDialog` appear
- Press Enter
--> The app creates a single task with a the pasted tasks structure as name, instead of confirming the dialog and creating multiple tasks

It seems that the input field still has focus and then triggers a single task creation. If you could fix this, I think we could merge the PR.

tests/javascript/unit/utils/textToTask.spec.js
```diff
@@ -0,0 +1,170 @@
...
+	it('should convert different prefixes', () => {
+		const text
+			= `
+task
+- task
++ task
+* task
+- [] task
+- [x] task
+- [X] task
```
raimund-schluessler, line 52: I was under the impression, that this would create already completed tasks. While the markdown would suggest that, I am not sure, it would make sense.

t-h-e:
> It seems that the input field still has focus and then triggers a single task creation. If you could fix this, I think we could merge the PR. I think it should be possible to programmatically focus the `Create Tasks` button in the dialog, e.g. on mount.

Actually the focus is automatically on the first button in the dialog. The issue is, at least that is what it seems to me, that the button already reacts on `keydown` event. If afterwards the focus is set again to the input field in the `cancel` or `close` event, the `keyup` event is fired afterwards, which leads to the creation of an empty task.

As the first button seems to have the focus after showing the dialog, I prevent `keydown` and `keyup` events for that button. Not nice, but I could not find a better solution.

tests/javascript/unit/utils/textToTask.spec.js
t-h-e: True, we could create completed tasks, but I'm not sure what's the use case for that?

t-h-e:
Actually you are right. Focusing `Create Tasks` button works 😄

raimund-schluessler:
Thanks a lot, it works a lot better now.

However, whether the task input field is cleared, depends on how the modal is closed. That might be a bit unexpected.

After creating the tasks, the input is:
- cleared when clicking `Close`
- not cleared when clicking on the `X` or outside the modal

After pasting text, but before creating tasks, the input is:
- not cleared when clicking `Cancel`
- cleared when clicking on the `X` or outside the modal


I would propose that the input field is always cleared after the tasks where created, but never cleared when canceling the task creation.

raimund-schluessler:
> @jancborchardt Could you please add @t-h-e to nextcloud/tasks?

@jancborchardt Ping :slightly_smiling_face:

t-h-e:
I forgot a `!` 🙈
See https://github.com/nextcloud/tasks/pull/2273/commits/55438dacce99a0d49afd029c9d2d317849fca088

t-h-e:
> I would propose that the input field is always cleared after the tasks where created, but never cleared when canceling the task creation.

Note that Firefox and Chrome behave differently. In Firefox it behaves now as you described. In Chrome, the pasted text is never placed in the input field.

raimund-schluessler (approved):

raimund-schluessler:
It works well now, I would say we can merge it.

> Note that Firefox and Chrome behave differently. In Firefox it behaves now as you described. In Chrome, the pasted text is never placed in the input field.

We might be able to fix this by explicitly setting the input field value to the pasted string after canceling.

brahimjarallah:
> Here is an example: BatchTaskCreation [ BatchTaskCreation ](https://user-images.githubusercontent.com/999552/239743809-c81e629e-0953-4829-9236-50c45a344ed1.gif) [ ](https://user-images.githubusercontent.com/999552/239743809-c81e629e-0953-4829-9236-50c45a344ed1.gif)

why this feature doesn't work for me in nextcloud 27 ?

raimund-schluessler:
Because, as I wrote in the issue you created, and as you can see in the milestone, it is unreleased. It will be released with v0.16.0.

RayBB:
@t-h-e thanks for your hard work on this. I'm so happy to see it finally released :)

jancborchardt:
@raimund-schluessler @t-h-e sorry for missing the ping – @t-h-e you are added to the Nextcloud org and the Tasks contributors now. :)

Issue #2268, opened 2023-05-15
Unable to sync tasks if web app is not used once 


avinash-0007:
### Steps to reproduce
1.Create an account on nextcloud
2. Now use  task [mobile app](https://f-droid.org/en/packages/org.tasks/) on mobile device
3. default **Task calendar** is missing if you have-not used task app once on web.

### Expected behaviour
Default task Calendar should be present on Task Mobile app otherwise sync is broken. 
App link: https://play.google.com/store/apps/details?id=org.tasks&pcampaignid=MKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1&pcampaignid=pcampaignidMKT-Other-global-all-co-prtnr-py-PartBadge-Mar2515-1&pli=1
I see that this is handled for calendar app 

https://github.com/nextcloud/server/blob/v25.0.4/apps/dav/lib/HookManager.php#L177 but we need this to be handled for task calendar.

### Actual behaviour
No default Task Calendar Present

### Server configuration

**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)
Nextcloud -25.0.6
**Tasks version:** (see Nextcloud apps page)
0.14.5

raimund-schluessler:
Closing in favor of https://github.com/nextcloud/calendar/issues/4554.

PR #2277, opened 2023-05-29
Fix for Unable to sync tasks if web app is not used once

avinash-0007:
This Pull request contains the fix for Unable to sync tasks if web app is not used once

https://github.com/nextcloud/tasks/issues/2268

### DIFF
--- a/lib/Migration/CreateTasksCalendar.php
+++ b/lib/Migration/CreateTasksCalendar.php
@@ -1,0 +1,161 @@
+<?php
+/**
+ * @copyright Copyright (c) 2023 Avinash Gusain <e.foundation>
+ *
+ * @author Avinash Gusain <avinashg@murena.io>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


raimund-schluessler:
I think this belongs into server. If you integrate this into the Tasks app, you won't be able to sync Tasks when the Tasks app is not installed, instead of not used once. Probably an improvement, but not really the idea solution of fix, I would say.

@nextcloud/calendar What do you think?

tcitworld:
Agreed with @raimund-schluessler, and in any case https://github.com/nextcloud/calendar/issues/4554#issuecomment-1534455585 will fix this (applies to new calendars in the Calendar app as well as the initial calendar created at the user's first login here: https://github.com/nextcloud/server/blob/45d7cee7f8220775b0229da5f5cd61fe7456c6c6/apps/dav/lib/HookManager.php#L172-L196

PR would be quite fine other wise, nice work.

lib/Migration/CreateTasksCalendar.php
```diff
@@ -0,0 +1,161 @@
...
+		}
+		//get all principal uri having no task calendar with component as TODO but have personal calendar
+		$result = $this->getPrincipalUriByCalendar();
+		foreach ($result as $row) {
+			$principal = $row['principaluri'];
+			$user = str_replace('principals/users/', '', $principal);
+			if ($this->userManager->userExists($user)) {
+				$taskUri = $this->getUniqueTaskUri($principal, self::TASKS_CALENDAR_URI);
+				$this->calDav->createCalendar($principal, $taskUri, [
+					'{DAV:}displayname' => self::TASKS_CALENDAR_NAME,
```
tcitworld, line 152: **nitpick:** This should be translated.

PR #2329, opened 2023-08-27
Add pr feedback action

Fenn-CS:


raimund-schluessler:
Will this later be updated from https://github.com/nextcloud/.github/blob/master/workflow-templates/pr-feedback.yml ?

PR #2346, opened 2023-09-22
Don't init router in dashboard.js

marcelklehr:
When the tasks app is enabled, the dashboard script automatically appends `#/` to the dashboard URL, because the router is initialized.


raimund-schluessler:
@marcelklehr Thanks for this PR, this is a good idea. However, I decided that the router shouldn't be used in store at all. So I refactored this a bit in #2362. This Supersedes this PR here.

marcelklehr:
Awesome, even better :)

PR #2436, opened 2024-01-03
feat: show link to Deck for deck cards

raimund-schluessler:
This adds a link to the Deck app for Deck cards and stacks.

@juliushaertl Am I correct that it's not necessary to check whether the Deck app is enabled, because having entries from Deck already indicates it is?


juliushaertl:
Yes, the calendars are only exposed if deck is enabled.

juliushaertl (approved):

Issue #2453, opened 2024-01-13
Wait for new release of nextcloud/vue


raimund-schluessler:
We need a new release of `nextcloud/vue` to finally fix #2366 and get the filter dropdown to work better.

Todo:
- Adjust import of `NcProgressBar` here https://github.com/nextcloud/tasks/blob/987f8b8bf565d543ba9fde1e0fbfa28f0ab5ba92/src/components/TaskBody.vue#L200

PR #2464, opened 2024-01-20
feat: use circular progress bar instead of linear bar

### LINKED ISSUE #1439: Length of completion bar is wrong
### Steps to reproduce
1. Go to the completed tasks
2. Depending on the symbols the blue bar is longer or shorter with same completed percentage
length_completed_bar

### Expected behaviour
The bar has in all cases where the completed percentage is the same the same length

### Actual behaviour
Depending on symbols (notes attached etc.) the bar length differs

### Server configuration

**Nextcloud version:** 20.0.8

**Tasks version:** 0.13.6

**Updated from an older Nextcloud or fresh install:** updated

### Client configuration
**Browser:** Firefox 86.0

**Operating system:** Ubuntu 20.10

raimund-schluessler:
This PR changes the progress / completion bar from linear to circular. It uses less space and the length and position is always the same. Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5100.

Before:
grafik

After:
grafik

Closes #1439.

### DIFF
--- a/src/components/NcProgressBar.vue
+++ b/src/components/NcProgressBar.vue
@@ -1,0 +1,196 @@
+<!--
+  - @copyright Copyright (c) 2020 Marco Ambrosini <marcoambrosini@icloud.com>
+  -
+  - @author Marco Ambrosini <marcoambrosini@icloud.com>
+  -
+  - @copyright Copyright (c) 2024 Raimund Schlüßler <raimund.schluessler@mailbox.org>
+  - @author Raimund Schlüßler
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


szaimen (approved):
LGTM from design perspective. Thanks for implementing this! :)

raimund-schluessler:
Merging now. Once there is a new nc/vue release, we will have to adjust the import accordingly. Tracked in #2453.

jancborchardt:
I find the linear one a more obvious for comparisons. :) Generally the rule of thumb is:
- Linear for longer loading and if space is available (which is both the case here)
- Circular for shorter and where space is small (e.g. spinners, or mobile app installation with indicator on icon)

But your call of course. :) As an enhancement, it would be nice if the indicator circles all line up, so if they are always the element left next of the action menu.

raimund-schluessler:
> I find the linear one a more obvious for comparisons. :)

I understand your point, and I also think that linear bars can have advantages here. But the problem was that the bars don't always have the same total length. And there isn't really a fix for this, since subtasks are indented and shorter. So comparing different bars could be a bit misleading, see the originial issue #1439.

So, I think cirular is a bit better suited here, although it's probably a gray area 🙈 

> As an enhancement, it would be nice if the indicator circles all line up, so if they are always the element left next of the action menu.

That's a good idea, I will do this. see #2466.

jancborchardt:
@raimund-schluessler ah, missed the link to the original issue, sorry. :) Right, the issue of variable widths (or rather start and end points) is indeed an issue. Then let's see how this fares. :)

PR #2480, opened 2024-02-01
feat: use web history mode

### LINKED ISSUE #2474: feat: switch to `webHistory` mode in vue router
We currently use `createWebHashHistory`  instead of `createWebHistory`, but this creates issues with the skip to navigation / content buttons.
So switching to `createWebHistory`  should solve this issue and make the skip buttons work. It also gives a nicer looking URL.

It would be nice to merge #1971 first, because it reduces the migration effort a bit.

raimund-schluessler:
This PR switches from `createWebHashHistory` to `createWebHistory`. This also fixes the skip-actions buttons, which did not work with `createWebHashHistory` mode.

A drawback is that current bookmarks to a certain collection or calendar won't work anymore and have to be recreated/adjusted.

Closes #2474.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -25,8 +25,16 @@ import AppSidebar from './views/AppSidebar.vue'
 import Calendar from './views/AppContent/Calendar.vue'
 import Collections from './views/AppContent/Collections.vue'
 
+import { getRootUrl, generateUrl } from '@nextcloud/router'
+
 import { h } from 'vue'
-import { createWebHashHistory, createRouter, RouterView } from 'vue-router'
+import { createWebHistory, createRouter, RouterView } from 'vue-router'
+
+const webRootWithIndexPHP = getRootUrl() + '/index.php'
+const doesURLContainIndexPHP = window.location.pathname.startsWith(webRootWithIndexPHP)
+const base = generateUrl('apps/tasks', {}, {
+	noRewrite: doesURLContainIndexPHP,
+})
 
 const routes = [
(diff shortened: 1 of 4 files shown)


src/router.js
```diff
@@ -25,8 +25,16 @@ import AppSidebar from './views/AppSidebar.vue'
...
+
 import { h } from 'vue'
-import { createWebHashHistory, createRouter, RouterView } from 'vue-router'
+import { createWebHistory, createRouter, RouterView } from 'vue-router'
+
+const webRootWithIndexPHP = getRootUrl() + '/index.php'
+const doesURLContainIndexPHP = window.location.pathname.startsWith(webRootWithIndexPHP)
+const base = generateUrl('apps/tasks', {}, {
+	noRewrite: doesURLContainIndexPHP,
+})
```
raimund-schluessler, line 37: @susnux @Antreesy I saw that other apps (e.g. Calendar, see https://github.com/nextcloud/calendar/blob/b93fed45c3c61faa82c7fe1f3aa27c4d3f3ee792/src/router.js#L39-L43) do it like this, but it seems a bit complicated to me. Is there an easier way?

Antreesy: Haven't seen anything like it. 
But you could check the latest release of nextcloud/router, getRootUrl was rewritten, perhaps, it covers that

raimund-schluessler: Thanks for the hint. I had a look, but it doesn't give exactly the same result. I guess it works as it is now. We can always improve later on.

Issue #1184, opened 2020-08-10
[Feature Request] Editable completed dates


freayd:
**Is your feature request related to a problem? Please describe.**
Completed dates are currently shown in the details footer tooltip as read-only data. Sometimes I mark tasks completed only a few days after having actually completed them, but I would like to be able to edit the completed date to reflect when it actually happened.

**Describe the solution you'd like**
It would be nice to be able to edit completed dates directly into Nextcloud without having to use another piece of software only for this task. It could be implemented as proposed in the issue #312. The field should probably only be shown if a task has been marked as completed. Here is the mockup originally proposed by @TruputiGalvotas:
![image](https://user-images.githubusercontent.com/19952587/54073913-3e0b5080-4295-11e9-9595-01aa4d63c4ef.png)

**Describe alternatives you've considered**
I am currently using Thunderbird to do this, but I'd rather stick to Nextcloud.

PR #2686, opened 2024-09-22
feat: allow editing completion date

creightonfrance:
- #1184 

This PR will allow changing the task completion date. 

- If user tries to set completion date in the future, they will see error dialog.
- If user deletes completion date, the task will marked as not completed.

### DIFF
--- a/src/components/AppSidebar/DateTimePickerItem.vue
+++ b/src/components/AppSidebar/DateTimePickerItem.vue
@@ -120,7 +120,7 @@ export default {
 		 */
 		checkOverdue: {
 			type: Boolean,
-			default: false,
+			default: true,
 		}
 	},
 	data() {
--- a/src/components/AppSidebar/DateTimePickerItem.vue
+++ b/src/components/AppSidebar/DateTimePickerItem.vue
@@ -115,6 +115,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Whether the date can be considered 'overdue'
+		 */
+		checkOverdue: {
+			type: Boolean,
+			default: false,
+		}
 	},
 	data() {
 		return {
@@ -139,7 +146,7 @@ export default {
 			return this.date.isValid()
 		},
 		isOverdue() {
-			return overdue(this.date)
+			return this.checkOverdue && overdue(this.date)
 		},
 	},
 	methods: {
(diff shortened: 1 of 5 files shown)


raimund-schluessler:
Seems to work fine. But I am not sure we should show this date so prominently. I would rather move it down into the details section. Probably below the `status` drop-down. What do you think?

creightonfrance:
That makes sense, it might be better under the status input since that determines this field's visibility. Making that change now 👍🏻

creightonfrance:
# Changes
I moved the field down to the details section. I also added a flag for the overdue check.

# Question
Could you help me out with local lint checks? I run `make` and  `npm run lint`, and I get hundreds of errors (mainly issues finding dependencies). I've fully opened up permissions locally for all files via `chmod 777 * -R`. Am I missing a step?

raimund-schluessler:
> # Question
> 
> Could you help me out with local lint checks? I run `make` and `npm run lint`, and I get hundreds of errors (mainly issues finding dependencies). I've fully opened up permissions locally for all files via `chmod 777 * -R`. Am I missing a step?

Unfortunately, I see the same locally. And I couldn't figure out yet, why that happens. Probably worth opening an issue.

src/components/AppSidebar/DateTimePickerItem.vue
```diff
@@ -115,6 +115,13 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
+		 * Whether the date can be considered 'overdue'
+		 */
+		checkOverdue: {
+			type: Boolean,
+			default: false,
```
raimund-schluessler, line 123: Could you please make the default `true` to keep the behavior the same?

creightonfrance:
> > # Question
> > Could you help me out with local lint checks? I run `make` and `npm run lint`, and I get hundreds of errors (mainly issues finding dependencies). I've fully opened up permissions locally for all files via `chmod 777 * -R`. Am I missing a step?
> 
> Unfortunately, I see the same locally. And I couldn't figure out yet, why that happens. Probably worth opening an issue.

I figured it out: issue only happens when the tasks repo is under the nextcloud/server repo (in apps, extra-apps, etc.). You have to run `make` in the base directory for nextcloud server, then `make` for this repo and lint should work fine 👍🏻 no code needed

raimund-schluessler (approved):
Seems to work fine. And sorry it took me so long to have a look.

Issue #1128, opened 2020-07-17
[Feature Request] Simplified "Priority" slider


Torenga:
RFC 5545 specifies 1-4 as High, 5 as Medium and 6-9 as high with 0 as "Undefined/No priority".

Most task managers, including all I've been able to sync with NC Tasks uses a four-level system; No Priority-Low-Medium-High instead of the full 10 levels defined in RFC 5545. This corresponds with NC Tasks own four colour scheme - Grey (No priority), Blue (Low priority), Yellow (Medium priority) and Red (High priority). The 10-level slider in Tasks is confusing when compared to _it's own colour scheme_ and other task managers.

I request that the priorities slider is either changed to the defacto standard, RFC 5545-compliant four-levels; No priority Grey=0, High priority Red=1, Medium priority Yellow=5 and Low priority Blue=9 - Or have the option to toggle this feature akin to Calendars "Simplified Editor".

Thank you for your consideration and providing a great NC feature!

raimund-schluessler:
I think it doesn't hurt to support all 10 priority levels specified by RFC 5545, but I don't have a too strong opinion on that. We only use four colors, since more would be hard to memorize and are not very meaningful. One still might like to have a finer priority setting than four. But if the majority of @nextcloud/designers likes to only have four levels I am fine with that.

Torenga:
> I think it doesn't hurt to support all 10 priority levels specified by RFC 5545

Part of my argument is that it's confusing and de facto unstandard, which I consider bad for the user experience if NC Tasks is synced with practically any other task manager.

There's also the possibility for both 4 and 10 levels with a "Simplified Editor" toggle. Looking forward to hear what other designers say. :)

freayd:
It might help if the colour of the star next to the number would change when the slider is moved. As I move the slider, it'd be nice to know if the result is going to red, yellow or blue. Otherwise, it's kind of a guessing game.

raimund-schluessler:
There was also another idea to change the priority handling in #1166. See my answer in https://github.com/nextcloud/tasks/issues/1166#issuecomment-673313979. Let's please use this issue here for discussing how to proceed with priority settings (since the ideas in #1128 and #1166 are a bit contradictory).

Torenga:
_Thank you @raimund-schluessler for recognizing that these two topics are related and bringing the discussion together._

## Simplifying use

[...]

PR #2687, opened 2024-09-23
feat: simplify priority handling

creightonfrance:
- #1128 

# Changes
- Sidebar will now present the priority value as a dropdown (High, Medium, Low)
  - These values are mapped accordingly:
    - High (9)
    - Medium (5)
    - Low (1)
    - No Priority (0)
- Priority changing from task list item goes through the following states: `0` => `1` => `5` => `9` => `0`

image

# TODO

- [ ] Unit test (holding off, anticipating change requests and file conflicts with #2686)


Felitendo:
Does this work with the tasks.org app? I feel like it's difficult to just change the priority system without any repercussion to mobile apps or am I wrong here?

raimund-schluessler:
The priority values we write to the CalDAV database are still the same. The mapping from these values to the levels we show to the user would just be different, or rather, we show less options to the user. So there is no problem with other apps here.

Felitendo:
Great, thanks for the heads up! Also thank you for working on the tasks app again. Looking forward to 0.17 👀

Issue #2181, opened 2023-01-24
Task's start/due date timestamp gets cut off


Insprill:
### Steps to reproduce
1. Create a task with a start or due date at a specific time.
2. Ensure your browser window is wide enough for the full date and time to show.
3. Observe the time being cut off.

### Expected behaviour
The full date should be visible if there's enough room.

### Actual behaviour
The timestamp gets cut off
![image](https://user-images.githubusercontent.com/52843537/214193804-228b6034-378d-479d-8a79-6bb0f3fd6948.png)

Similarly, the date switches to the short version a little prematurely when the browser window is small, due to the menu on the left being collapsed allowing for more space.
![image](https://user-images.githubusercontent.com/52843537/214193694-9db47597-98c3-4427-ab95-1b2339f1401f.png)

### Server configuration
**Operating system**: Ubuntu 22.04

**Web server:** Nginx

**Database:** MySQL

**Nextcloud version:** 25.0.2

**Tasks version:** 0.14.5

**Updated from an older Nextcloud or fresh install:** Updated from 24.0.8

```
No errors have been found.
```

**Nextcloud configuration:**

```
<?php
$CONFIG = array (
  'apps_paths' =>
  array (
    0 =>
    array (
      'path' => '/snap/nextcloud/current/htdocs/apps',
      'url' => '/apps',
      'writable' => false,
    ),
    1 =>
    array (
      'path' => '/var/snap/nextcloud/current/nextcloud/extra-apps',
      'url' => '/extra-apps',
      'writable' => true,
    ),
  ),
  'supportedDatabases' =>
  array (
    0 => 'mysql',
  ),
  'memcache.locking' => '\\OC\\Memcache\\Redis',
  'memcache.local' => '\\OC\\Memcache\\Redis',
  'redis' =>
  array (
    'host' => '/tmp/sockets/redis.sock',
    'port' => 0,
  ),
  'log_type' => 'file',
  'logfile' => '/var/snap/nextcloud/current/logs/nextcloud.log',
  'logfilemode' => 416,
  'instanceid' => '[redacted]',
  'passwordsalt' => '[redacted]',
  'secret' => '[redacted]',
  'trusted_domains' =>
  array (
    0 => '[redacted]',
  ),
  'datadirectory' => '[redacted]',
  'dbtype' => 'mysql',
  'version' => '25.0.2.3',
  'overwrite.cli.url' => '[redacted]',
  'dbname' => '[redacted]',
[...]

PR #2833, opened 2025-04-06
Auto determine task date width on large screens

asavageiv:
Fixes https://github.com/nextcloud/tasks/issues/2181


asavageiv:
What is your worst case test so I can make sure that works too.

What we want is for the task text to take priority until the other data
reaches its min-width, right?

jaylinski:
I would suggest a max-width-[ellipsis](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) with a [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)-attribute.

PR #2846, opened 2025-04-27
Group tasks with same due date

pieterdd:
Here's a draft of something I've wanted for a while:

image

Current shortcomings:
- UI freezes when I click one of the tasks (haven't figured out a solution yet, but I bet it's related to the way I update calendar objects in the `filteredCalendars` computation)
- This grouping should only apply when due date sorting is active

Is this a feature you'd be willing to add to Tasks?


raimund-schluessler:
What is the difference to the `Week` view? The `Week` view already groups tasks by their due date, so I don't understand what would be different here.

pieterdd:
You're right :exploding_head: Not sure why I didn't see that before, but that's exactly what I need.

PR #2931, opened 2025-09-26
fix: show modal when date related to a reminder is deleted

### LINKED ISSUE #2751: fix: deleting the date a reminder is related to should delete the reminder
@jaylinski We should stay in accordance to the CalDAV specification, which states that relative alarms without the date the alarm is related to are not allowed. See the description in https://icalendar.org/iCalendar-RFC-5545/3-6-6-alarm-component.html:
```
In an alarm set to trigger on the "START" of an event or to-do,
      the "DTSTART" property MUST be present in the associated event or
      to-do.  In an alarm in a "VEVENT" calendar component set to
      trigger on the "END" of the event, either the "DTEND" property
      MUST be present, or the "DTSTART" and "DURATION" properties MUST
      both be present.  In an alarm in a "VTODO" calendar component set
      to trigger on the "END" of the to-do, either the "DUE" property
      MUST be present, or the "DTSTART" and "DURATION" properties MUST
      both be present.
```

So, the options 2 and 3 in your list are not possible by the specification alone. Furthermore, the user would still see the alarms being present, and might falsely assume that they are still working, where in reality they won't.

However, we are free to convert relative alarms to absolute alarms, if the related date is deleted. We could present the user with a popup to let him decide whether he wants to:
- cancel the deletion of due or start date
- convert relative alarms related to the deleted date to absolute alarms
- delete the relative alarms related to the deleted date

The other option would be to not confront the user with this problem, and simply convert relative to absolute alarms. This is what we mostly do, to prevent pop-ups and dialogs as they often interfere with the workflow.

@jancborchardt and @nimishavijay might also have an opinion here.

_Originally posted by @raimund-schluessler in https://github.com/nextcloud/tasks/issues/2743#issuecomment-2576992422_

jaylinski:
According to the CalDAV specification, relative alarms without the date the alarm is related to are not allowed. In order to comply with the spec, we ask the user if the reminder should be converted to an absolute date or if it should be discarded.

Fixes #2751


raimund-schluessler:
Thanks for starting to work on this, this is the last issue I would like to fix before releasing 0.17.0.

Since I saw your TODO comment: I would say the date should only be deleted after the user decides whether to keep or delete the alarm. If the user select `cancel`, the date should be kept.

jaylinski:
> Thanks for starting to work on this, this is the last issue I would like to fix before releasing 0.17.0.
> 
> Since I saw your TODO comment: I would say the date should only be deleted after the user decides whether to keep or delete the alarm. If the user select `cancel`, the date should be kept.

@raimund-schluessler I agree. The date won't be kept, but restored if the user cancels the modal. Not perfect, but it was the "easiest" way to implement it. See implementation notes above.

PR is ready for review.

raimund-schluessler:
Thanks a lot for tackling this. I would say the behavior is ok for now. We can still improve it later.
Unfortunately, I won't have time to review it until the weekend.

raimund-schluessler:
The problem with this approach is also, that between deleting the date and answering the modal, we have an invalid task on the server which has a reminder but no related date. I guess in order to get Tasks 0.17.0 out we can accept that for now, but should take care of this soon.

PR #3027, opened 2026-01-19
Replace dependency on moment with dayjs

asavageiv:
Following nextcloud/server#57503, this saves about 15% of the zipped bundle size for Tasks.

### DIFF
--- a/tests/javascript/unit/store/tasks.spec.js
+++ b/tests/javascript/unit/store/tasks.spec.js
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach } from 'vitest'
+import { describe, it, expect } from 'vitest'
 import { createStore } from 'vuex'
 import { loadICS } from '../../../assets/loadAsset.js'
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
Thanks for the PR. Unfortunately, some unit tests related to moment now fail.

asavageiv:
I fixed the tests, but I think this needs more work. One difference between Moment and DayJS is that DayJS dates are immutable. There means that where before this would have worked:
```javascript
due.add(start.diff(currentstart), 'ms')
```
it now needs to be
```javascript
due = due.add(start.diff(currentstart), 'ms')
```

It doesn't look like the tests cover all these cases, so I will have to add some more tests and work my way through them manually.

asavageiv:
Ok, I think this is good to go now.

src/utils/dateStrings.js
raimund-schluessler: This needs to be adjusted later.

PR #3021, opened 2026-01-09
Support for recurring tasks

schulm:
Fixes #34

PR #3190, opened 2026-06-29
fix(TaskBody): show due date instead of start date in task list

### LINKED ISSUE #3187: regression: TaskBody shows start date instead of due date if start date is set
### Steps to reproduce

Using tasks

### Expected behavior

After downgrading to version 0.17.1, it works properly again. 
It seems that this only applies to entries with a start date.

Image

### Actual behaviour

In version 0.18.0, the start date is shown instead of the chosen due date.

Image

### Tasks app version

v0.18.0

### CalDAV-clients used

DAVx5

### Browser

Brave, Floorp

### Client operating system

Arch Linux

### Server operating system

_No response_

### Web server

None

### Database engine version

None

### PHP engine version

PHP 8.2

### Nextcloud version

34

### Updated from an older installed version or fresh install

None

### List of activated apps

```Text

```

### Nextcloud configuration

```Text

```

### Web server error log

```Text

```

### Log file

```Text

```

### Browser log

```Text

```

### Additional info

_No response_

schulm:
### Problem

Since the recurring-tasks change (#3021), the task list started showing the **start date** instead of the **due date** for tasks that have both set — a regression from 0.17.1 (#3187).

### Background

I tried to remember the reasoning behind the original change. As best I can reconstruct it, the idea was to surface *some* date for tasks that have only a start date (which previously showed nothing) — but the precedence ended up backwards, so the start date was preferred in **all** cases, hiding the due date whenever both were set.

Thinking it through again: the start date isn't the actionable date in the list, and recurrence doesn't change that — the due date is what matters, and it's also the value the recurrence completion advances (`task.due || task.start`). Putting a start date in the same unlabeled column as due dates was also the source of the "which date is this?" confusion.

### Fix

Revert the date column to its pre-0.18.0 behavior: **always show the due date.**

- `v-if="task.due || task.completed"`
- overdue highlight based on the due date only
- `dueDateShort` / `dueDateLong` use the due date directly

The recurring indicator icon is unaffected. Start-only tasks show no date text (the long-standing behavior) and still get the existing `CalendarClock` icon hint.

### Testing

- Verified the three date combinations (start+due → due, due-only → due, start-only → no date) on a running instance.
- `npm run lint`, `npm run dev`, and the unit test suite (`vitest`) all pass.

Fixes #3187

### DIFF
--- a/src/components/TaskBody.vue
+++ b/src/components/TaskBody.vue
@@ -82,7 +82,7 @@ License along with this library. If not, see <http://www.gnu.org/licenses/>.
 					:title="t('tasks', 'Task has a note')"
 					@click="openAppSidebarTab($event, 'app-sidebar-tab-notes')"
 					@dblclick.stop="openAppSidebarTab($event, 'app-sidebar-tab-notes', true)" />
-				<div v-if="task.start || task.due || task.completed" :class="{'date--overdue': overdue(task.startMoment || task.dueMoment) && !task.completed}" class="date">
+				<div v-if="task.due || task.completed" :class="{'date--overdue': overdue(task.dueMoment) && !task.completed}" class="date">
 					<span class="date__short" :class="{ 'date__short--completed': task.completed }">{{ dueDateShort }}</span>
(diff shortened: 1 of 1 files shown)


raimund-schluessler:
@schulm Thanks a lot for the comprehensive explanation and for providing the fix!

